PR #5, opened 2021-09-25
Add github workflow

fracek:
I added a workflow that:

 * Tests the project on multiple Python versions
 * Builds and checks the Python package
 * (On new tags only) Uploads the package to PyPi

You need to set the `pypi_token` secret to upload to PyPi.


martriay:
Thanks again!

PR #8, opened 2021-09-26
Include submodule with urllib

perama-v:
This resolves the error: "ModuleNotFoundError: No module named 'urllib.requests'" on first installation.


martriay:
Thanks a lot!

PR #12, opened 2021-09-28
Compile cmd. Contracts dir subdirectories lookup

marcellobardus:


marcellobardus:
That was tested locally, by creating a `contracts` dir in the root and creating `sub` subdirectory with `test.cairo` inside.
As expected two artifacts were created.

martriay:
Thanks!

Issue #6, opened 2021-09-25
Big picture view


fracek:
I've spent some more time thinking how to organize cairo projects and what I'd like to see. I think we should leverage the python ecosystem as much as possible:

 * `nile init` could use cookiecutter to create the project
 * `pypi` to distribute cairo packages
 * users should be able to configure nile (for example the contracts directory) from their `setup.cfg` file like other python tools

I think cairo projects should use the same folder structure as python projects, this has the added benefit that library developers can distribute the cairo files together with an sdk to interact with it (like `@uniswap/core` and `@uniswap/sdk`).

```txt
.
├── src
│  └── cairoswap
│     ├── __init__.py
│     ├── cairo
│     │  ├── __init__.py
│     │  ├── pool.cairo
│     │  └── router.cairo
│     ├── pool.py
│     └── router.py
├── tests
│  ├── pool.test.cairo
│  ├── test_pool.py
│  └── test_router.py
├── pyproject.toml
├── setup.cfg
└── tox.in
```

Developers can then create a tox task than runs python tests with pytest and cairo tests with `nile test`.

If users don't need/want to setup a full blown project they can revert to a minimal setup and use nile to build and test.

```txt
.
└── src
   ├── pool.cairo
   └── pool.test.cairo
```

perama-v:
I like it

martriay:
Like the approach as well, and even though I'm not a python expert I lean towards minimizing dependencies until strictly necessary. Do we _need_ cookiecutter? Why not a lightweight approach like a few `mkdir` and a single opinionated project structure? I'm sure there are good reasons but I'm unaware.

As of the pythonic project structure thing, I'm undecided

* On one hand, I think Cairo/StarkNet needs to improve their developer experience by like 100x yet, and having basic unopinionated tooling infra is key
* But on the other hand kickstarting the ecosystem with a single opinionated language might be a good strategy to have something working asap, and iterate on the way. Ethereum did ok just with Solidity, and truffle for a while until hardhat took on. So if you have to pick a dev env to work in Cairo today, that's python due to the resemblance to Cairo and because of the Cairo tooling (e.g. StarkNet's testing framework).

That's why I'm ok with going with pythonic but probably design with a multi-language/multi-stack future in mind.

[...]

PR #18, opened 2021-10-03
Add Makefile to init template

fracek:
As mentioned in #6, I added a Makefile to the init template.

I also changed the filename of the test file so that pytest collects it automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,10 +33,8 @@ nile init
 ✅ Dependencies successfully installed
 🗄  Creating project directory tree
 ⛵️ Nile project ready! Try running:
-
-nile compile
 ```
-This command creates the project directory structure and installs `cairo-lang`, `pytest`, and `pytest-asyncio` for you.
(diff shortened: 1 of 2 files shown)


martriay:
Thanks! Looks good, would you mind adding some Makefile documentation on the README?

martriay (approved):

README.md
```diff
@@ -33,10 +33,8 @@ nile init
 ✅ Dependencies successfully installed
 🗄  Creating project directory tree
 ⛵️ Nile project ready! Try running:
-
-nile compile
```
fracek, line 37: I think this was a runaway line. Let me know if it was there for a reason.

fracek:
I added a line to the readme.

README.md
martriay: it was meant to work as an example command you can run with nile (notice the colon in the line above)

PR #20, opened 2021-10-13
feat: adding nile version command

### LINKED ISSUE #19: add `version` command
Consider using [click](https://click.palletsprojects.com/en/7.x/api/?highlight=version_option#click.version_option)

milancermak:
This PR adds a `nile version` command. Resolves #19. Example:

```
$ nile version
0.0.13.post1.dev3+gc9876f6.d20211013
```

Let me know if you want to format the output differently or add something (e.g. cairo compiler version).

It took me a while to understand how the project is set up, it might be that I overlooked something. Please let me know if that's the case. However, if I understand it correctly, it's the `setuptools_scm` that's in charge of setting the version string. This PR should be able to work with that seamlessly.

I have two suggestions:

1) Have you though about dropping support for py3.7? That way I wouldn't have to do all the try-except for importlib. Python 3.8 already ships with importlib.

2) Click has a [`version_option`](https://click.palletsprojects.com/en/7.x/api/?highlight=version_option#click.version_option) decorator that adds support for running commands with `--version`. The style is different than having version as a standalone command, but both `python` and the cairo toolchain use this convention. What do you think about using only `--version`?

### DIFF
--- a/src/nile/commands/version.py
+++ b/src/nile/commands/version.py
@@ -1,7 +1,7 @@
-"""Command to print Nile version"""
+"""Command to print Nile version."""
 from nile import __version__ as nile_version
 
 
 def version_command():
-    """Print Nile version"""
+    """Print Nile version."""
     print(nile_version)


src/nile/commands/version.py
```diff
@@ -0,0 +1,9 @@
+try:
```
fracek, line 1: Need a module docstring to silence black :)

PR #26, opened 2021-11-04
Add `base_project` directory and update to Cairo 0.5.0

martriay:
This PR changes how `init` bootstraps a new project and also updates the sample tests and contracts to work with the latest Cairo release.

PR #27, opened 2021-11-04
Integrate starknet devnet

### LINKED ISSUE #23: integrate starknet-devnet
https://github.com/Shard-Labs/starknet-devnet/tree/master

martriay:
Fixes #23. Depends on #26.

- [x] install it on `nile install`
- [x] add `deploy` method
- [x] handle deployments (simple address:artifact mapping)
- [x] add `invoke` method
- [x] add `call` method
- [x] add `node` command
- [x] update readme
- [x] prepare [blogpost](https://medium.com/@martriay/manage-your-starknet-deployments-with-nile-%EF%B8%8F-e849d40546dd)

### DIFF
--- a/src/nile/commands/node.py
+++ b/src/nile/commands/node.py
@@ -1,0 +1,7 @@
+"""Command to start StarkNet local network."""
+import subprocess
+
+
+def node_command():
+    """Start StarkNet local network."""
+    subprocess.check_call("starknet-devnet")
--- a/src/nile/commands/node.py
+++ b/src/nile/commands/node.py
@@ -4,4 +4,10 @@ import subprocess
 
 def node_command():
     """Start StarkNet local network."""
-    subprocess.check_call("starknet-devnet")
+    try:
+        subprocess.check_call("starknet-devnet")
+    except FileNotFoundError:
+        print("")
+        print("😰 Could not find starknet-devnet, are you sure it is installed? Try with:\n")
+        print("   pip install starknet-devnet")
+        print("")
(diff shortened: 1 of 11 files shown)


src/nile/commands/init.py
```diff
@@ -25,95 +26,9 @@ def init_command():
 
     # create project directories
     print("🗄  Creating project directory tree")
-    create_contracts()
-    create_tests()
-
-    with open("Makefile", "w") as fp:
-        fp.write(makefile)
+    copy_tree(os.path.join(os.path.dirname(__file__), "../base_project"), ".")
```
fracek, line 29: Maybe using `pathlib.Path` would make this cleaner? E.g. `Path(__file__).parent / 'base_project')`

martriay: thanks! will fix this in #26

src/nile/commands/node.py
```diff
@@ -0,0 +1,7 @@
+"""Command to start StarkNet local network."""
+import subprocess
+
+
+def node_command():
+    """Start StarkNet local network."""
+    subprocess.check_call("starknet-devnet")
```
fracek, line 7: Should it catch `FileNotFoundError` and recommend to install `starknet-devnet` using `nile install`?

fracek (commented):
I believe calling the current test network `mainnet` will be confusing once main net will be deployed at the end of the month. I suggest changing the usage of `mainnet` to `testnet`.

martriay:
> I believe calling the current test network `mainnet` will be confusing once main net will be deployed at the end of the month. I suggest changing the usage of `mainnet` to `testnet`.

Do you think so? I thought the goerli deployment behaved as mainnet as of today, such that if in the future you use the `starknet` command with these params, it will send txs to mainnet. I was under the impression goerli is just the settlement layer but the starknet nodes were already "the main network".

fracek:
That's true, we can update the names as they come. Then it all looks good to me!

PR #33, opened 2021-11-18
Added support for signer

exp-table:


src/nile/commands/proxy.py
```diff
@@ -0,0 +1,68 @@
...
+from nile.signer import Signer
+
+def proxy_setup_command(signer, network):
+    """Deploy an Account contract for the given private key."""
+    signer = Signer(int(os.environ[signer]))
+    if accounts.exists(str(signer.public_key), network):
+        signer_data = next(accounts.load(str(signer.public_key), network))
+        signer.account = signer_data["address"]
+        signer.index = signer_data["index"]
+    else: #doesn't exist, havbe to deploy
```
fracek, line 20: There should be a space between `#` and `doesn't`. There's a type (`havbe` should be `have`).

src/nile/commands/proxy.py
```diff
@@ -0,0 +1,68 @@
...
+        signer.index = accounts.current_index(network)
+        subprocess.run(f"nile deploy Account {signer.public_key} --alias account-{signer.index}", shell=True)
+        address, _ = next(deployments.load(f"account-{signer.index}", network))
+        #initialize account
```
fracek, line 24: Space between `#` and `initialize`

fracek:
I like this! I added a couple of nitpicky comments.

exp-table:
> I like this! I added a couple of nitpicky comments.

Fixed it!

fracek:
Ok there are a couple of linting issues. My advice is to run `tox -e format` to automatically format everything, then `tox -e lint` to double check everything is formatted correctly.

martriay:
This looks very nice! Do you think `proxy` could be merged into `send`? Or maybe I'm not understanding the different use case. In any case, I don't think that's the right name for it since it refers to a different kind of contracts which delegate execution to other contracts. Maybe `raw_execute`?

Future work could be:

1. auto generating a few accounts on `nile node`, have a default one and `--account` to select by index or alias
2. generating a private key, needed for `1` and pretty useful

src/nile/commands/proxy.py
```diff
@@ -0,0 +1,68 @@
...
+def proxy_setup_command(signer, network):
+    """Deploy an Account contract for the given private key."""
+    signer = Signer(int(os.environ[signer]))
+    if accounts.exists(str(signer.public_key), network):
+        signer_data = next(accounts.load(str(signer.public_key), network))
+        signer.account = signer_data["address"]
+        signer.index = signer_data["index"]
+    else: # doesn't exist, have to deploy
+        signer.index = accounts.current_index(network)
+        subprocess.run(f"nile deploy Account {signer.public_key} --alias account-{signer.index}", shell=True)
```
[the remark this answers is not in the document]
exp-table: I didn't even think about it when prototyping initially. Good catch, fixed it!

exp-table:
> This looks very nice! A few comments:
> 
>     * Do you think `proxy` could be merged into `send`? Or maybe I'm not understanding the different use case. In any case, I don't think that's the right name for it since it refers to a different kind of contracts which delegate execution to other contracts. Maybe `raw_execute`?
> 
>     * Could you expand the documentation to include an example env file and commands snippets with actual parameters? That help users a lot.
> 

I renamed them to `send` and `raw-execute`. As you said, `proxy` isn't the right label for what it does.
I also added an example `example.env` file as well as enhanced documentations for `send` and `raw-execute`.

fracek (approved):
Looks good to me. All issues have been addressed, thanks!

src/nile/signer.py
```diff
@@ -0,0 +1,48 @@
+"""Utility for sending signed transactions to an Account on Starknet."""
+import subprocess
+
+from starkware.cairo.common.hash_state import compute_hash_on_elements
```
fracek, line 4: This should be be wrapped in a try catch to avoid throwing when users don't have cairo installed yet (and want to use nile to install it).

fracek:
One last thing before I'm ready to merge. Sorry for letting you wait so long to review this PR!

README.md
```diff
@@ -88,6 +88,56 @@ A few things to notice here:
...
+🌕 artifacts/Account.json successfully deployed to 0x07db6b52c8ab888183277bc6411c400136fe566c0eebfb96fffa559b2e60e794
+📦 Registering deployment as account-0 in localhost.deployments.txt
+Invoke transaction was sent.
+Contract address: 0x07db6b52c8ab888183277bc6411c400136fe566c0eebfb96fffa559b2e60e794
+Transaction hash: 0x17
+```
+
+A few things to notice here:
+
+1. `nile set <env_var>` looks for an environement variable with the same name whose value is a private key
```
martriay, line 107: there's no `nile set` command :)

martriay (changes requested):

README.md
```diff
@@ -88,6 +88,56 @@ A few things to notice here:
 3. The `--alias` parameter lets me create an unique identifier for future interactions, if no alias is set then the contract's address can be used as identifier
 4. By default Nile works on local, but you can pass `--network mainnet` to deploy directly to a public chain! Notice that `mainnet` refers to StarkNet main chain, that's settled on Goerli testnet of Ethereum ([mainnet deployment this month!](https://medium.com/starkware/starknet-alpha-is-coming-to-mainnet-b825829eaf32))
 
+### `setup`
+You can find an exemple `.env` file in `example.env`. These are private keys only to be used for testing and never in production.
```
fracek, line 92: `exemple` -> `example`

src/nile/signer.py
```diff
@@ -0,0 +1,55 @@
...
+        self.index = 0
+
+    def sign(self, message_hash):
+        """Sign a message hash."""
+        return sign(msg_hash=message_hash, priv_key=self.private_key)
+
+    def get_nonce(self):
+        """Get the nonce for the next transaction."""
+        nonce = subprocess.check_output(
+            f"nile call account-{self.index} get_nonce", shell=True, encoding="utf-8"
```
fracek, line 33: Here it's missing a `--network` argument so when using a non-default network this call fails

fracek (changes requested):
Looks like it works well on a local network but does not work on mainnet because the network argument is missing in the call to get the nonce.

fracek (approved):
I run this in a fresh venv and everything is working as expected

martriay (approved):

martriay:
Thanks all 🎉

PR #38, opened 2021-12-12
Change command output on compilation error

### LINKED ISSUE #37: Report compile errors better
Hey,

I'd like to propose a change, but this might be a nitpick so feel free to ignore.

What do you think about improving the UX by showing a different failure emoji rather than the green thick mark when `nile compile` is done but the contract compilation fails? 

So instead of this:

```
nile compile contracts/contract.cairo
🔨 Compiling contracts/contract.cairo
contracts/contract.cairo:19:5: Expected exactly 0 expressions, got 1.
    return (0)
    ^********^
✅ Done
```

we could have 
```
nile compile contracts/contract.cairo
🔨 Compiling contracts/contract.cairo
contracts/contract.cairo:19:5: Expected exactly 0 expressions, got 1.
    return (0)
    ^********^
🛑 Compilation failed with 1 error
```

It's easier to spot that something actually failed (red vs. green). Would be also useful for a project with a lot of contracts, where the compilation failure might scroll off screen.

I also think if the compilation fails, `nile compile` should [exit with a non zero value](https://bash.cyberciti.biz/guide/The_exit_status_of_a_command) so that it can be used in scripting or CI workflows or in similar context.

I'm happy to send a PR if you believe doing this is worthwhile.

milancermak:
Partially resolves #37. If there's a compilation error, `nile compile` now outputs this:

```
$ nile compile contracts/contract.cairo
🔨 Compiling contracts/contract.cairo
contracts/contract.cairo:19:5: Expected exactly 0 expressions, got 1.
    return (0)
    ^********^
🛑 Failed to compile 1 contract
```

I ran tox and it reported this is good to go, hopefully I didn't forget anything else.

The return value of `nile compile` is still 0 even on compilation failure. I'd rather do that part in a separate PR and on all commands, since I think it makes sense.

Happy to hear your thoughts.

### DIFF
--- a/src/nile/commands/compile.py
+++ b/src/nile/commands/compile.py
@@ -25,7 +25,8 @@ def compile_command(contracts):
         all_contracts = get_all_contracts()
 
     results = [_compile_contract(contract) for contract in all_contracts]
-    failures = sum(r != 0 for r in results)
+    failed_contracts = [c for (c, r) in zip(all_contracts, results) if r != 0]
+    failures = len(failed_contracts)
 
     if failures == 0:
         print("✅ Done")
@@ -33,7 +34,9 @@ def compile_command(contracts):
         exp = f"{failures} contract"
         if failures > 1:
             exp += "s"  # pluralize
-        print(f"🛑 Failed to compile {exp}")
+        print(f"🛑 Failed to compile the following {exp}:")
+        for contract in failed_contracts:
+            print(f"   {contract}")
 
 
 def _compile_contract(path):
--- a/src/nile/commands/compile.py
+++ b/src/nile/commands/compile.py
@@ -18,17 +18,22 @@ def compile_command(contracts):
         print(f"📁 Creating {ABIS_DIRECTORY} to store compilation artifacts")
         os.makedirs(ABIS_DIRECTORY, exist_ok=True)
 
+    all_contracts = contracts
+
     if len(contracts) == 0:
         print(f"🤖 Compiling all Cairo contracts in the {CONTRACTS_DIRECTORY} directory")
-        for path in get_all_contracts():
-            _compile_contract(path)
-    elif len(contracts) == 1:
-        _compile_contract(contracts[0])
-    else:
-        for path in contracts:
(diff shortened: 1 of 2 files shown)


milancermak:
Ah, still need to `tox -e lint`

brb

src/nile/commands/compile.py
```diff
@@ -18,17 +18,22 @@ def compile_command(contracts):
...
+        all_contracts = get_all_contracts()
 
-    print("✅ Done")
+    results = [_compile_contract(contract) for contract in all_contracts]
+    failures = sum(r != 0 for r in results)
+
+    if failures == 0:
+        print("✅ Done")
+    else:
+        exp = f"{failures} contract"
```
martriay, line 33: is there an easy way to get _which ones_?

martriay (approved):
Looks really good! Left a minor suggestion, you can merge it without it :)

src/nile/commands/compile.py
milancermak: Yeah, I thought about that as well. It's easy to get it done. What do you think about an output like this?

```
$  nile compile
🤖 Compiling all Cairo contracts in the contracts directory
🔨 Compiling contracts/app.cairo
🔨 Compiling contracts/contract2.cairo
contracts/contract2.cairo:22:5: Expected exactly 0 expressions, got 1.
    return (2)
    ^********^
🔨 Compiling contracts/contract.cairo
contracts/contract.cairo:19:5: Expected exactly 0 expressions, got 1.
    return (0)
    ^********^
🛑 Failed to compile the following 2 contracts:
   contracts/contract2.cairo
   contracts/contract.cairo
```

It's, of course, easy to adjust, so if you have any input, I can make it happen ;)

martriay: that looks perfect!

milancermak: Done in the latest commit f04563d

milancermak:
I don't have access to merge it to `main`, you'll have to do it 😇

martriay:
Thanks 🎉

PR #39, opened 2021-12-19
Add mainnet compatibility

martriay:


fracek (approved):
LGTM.

Only comment was about the name `mainnet`/`goerli` but we discussed it last time and I think it's fine.

Issue #13, opened 2021-09-28
test nile commands


martriay:
it'd be nice to implement some basic testing, probably starting with `compile` and `clean`.
this would make reviews and releases much more easy.

TimNooren:
Hi @martriay, I'd love to contribute here. Would https://github.com/OpenZeppelin/nile/compare/main...TimNooren:add_tests?expand=1 be up to standard? If so I'd be happy to open a PR:)

martriay:
Yes! Please do :)

PR #40, opened 2021-12-20
Add tests for "compile" and "clean" CLI commands

TimNooren:
See https://github.com/OpenZeppelin/nile/issues/13

- Extracts `clean` command implementation to make it easier to test.
- Contract compilation itself is not tested.

Looks like I cannot assign reviewers, but as this is not time critical I'm just leaving it here for now:)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# vscode project settings
+.vscode/
--- a/README.md
+++ b/README.md
@@ -192,11 +192,17 @@ Nile uses tox to manage development tasks, you can get a list of
 available task with `tox -av`.
 
  * Install a development version of the package with `python -m pip install .`
- * Run tests with `tox`
  * Build the package with `tox -e build`
  * Format all files with `tox -e format`
  * Check files formatting with `tox -e lint`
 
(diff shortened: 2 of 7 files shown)


martriay:
Thanks for this very much needed PR! I hope it's the kickoff of a comprehensive test suite :)

How would you suggest we test compilation itself?

fracek:
I was thinking one way to test that is to have end to end tests in the CI. Basically have a shell where it runs all nile commands like a user would.

TimNooren:
Some end-to-end tests sounds like a good approach👍  Happy to give that a go (in a separate PR?). Could be done with `pytest` as well, since `click` has quite nice testing support.

fracek:
I too think end to end tests should be a separate PR. I will open an issue to track that.

martriay:
Agree with all your comments, let's continue that chat on the other issue. Really good work on this PR!

Could you add some testing instructions to the README? :)

TimNooren:
> Could you add some testing instructions to the README? :)

Added a bit more information. Let me know what you think:)

TimNooren:
Fixed the failing tests (I think). Missed one case where `starknet-compile` was being called.

martriay:
Thanks! 🎉

PR #42, opened 2021-12-23
Handle small tx hash on localhost

perama-v:
When deploying to localhost, the transaction 'hash' is small (`0x4`) causing the regex to skip the value leading to an unpack error


martriay:
thank you!

PR #43, opened 2021-12-23
Prevent truncation of 2 trailing characters

perama-v:
Error: `nile invoke xyz` results in `UNINITIALIZED_CONTRACT`.

Issue: When storing the contract address in `*.deployments.txt` the trailing two characters are omitted. When calling invoke on a contract verified to be deployed (e.g., on Voyager, address `0x123`), the contract printed to the console has two leading  0's: `0x00123`. The length of the address (excluding 0x prefix) appears to be 64 elsewhere (StarkNet CLI, Voyager).

Resolution: Increase the max length of the regex.


martriay:
Thanks!

Issue #41, opened 2021-12-21
End to end tests


fracek:
We discussed both on Discord and in #40 about adding end to end tests. These tests would work as follows (let me know if I missed any step):

* Create a new virtual environment
* Install nile
* Install cairo using `nile install`
* Initialize a new repository with `nile init`
* Compile the project
* Run tests
* Setup an account for the user (note: requires a private key)
* Deploy smart contract
* Use the account to interact with the smart contract
* Read value back from smart contract

This series of commands tests everything available in nile except `nile node`, which could be tested by running all steps that involve a network interaction twice (with local node and with test network).

Since StarkNet does not have transaction fees for now we can generate a new private key every time we run the tests and then let the account rot (we can't destroy it anyway).

Any thoughts?

martriay:
Sounds good! For the private key step we can just hardcode any key, even a dumb number like `123`.

The deployment step worries me because it depends on external factors (`starknet-devnet` on local, not to mention if we want to test public nets) but I guess that's fair since this is e2e not unit testing.

fracek:
I mentioned a random key because if we deploy on the alpha goerli network we need to deploy a new account every time. Not needed for local nodes since they reset state every time we run the tests.

TimNooren:
Really like the proposal :thumbsup: Some (general) questions/suggestions:

- I'm curious about the reason for having `nile install` to install `cairo-lang` and `starknet-devnet`, instead of including those in the installation requirements. The second option would avoid needing tests (that set up a virtual environment etc), but I'm sure there's a design decision that I'm not aware of:)

- `click` supports testing CLI commands using the [CliRunner](https://click.palletsprojects.com/en/8.0.x/api/#click.testing.CliRunner). Not 100% sure if the proposal is to run shell commands, but I think the CliRunner is at least a good first step.

- Regarding testing contract deployment and interactions: we could start with deploying to a local devnet, managed through `pytest` fixtures or something like docker-compose.

martriay:
> I'm curious about the reason for having nile install to install cairo-lang and starknet-devnet, instead of including those in the installation requirements. The second option would avoid needing tests (that set up a virtual environment etc), but I'm sure there's a design decision that I'm not aware of:)

The reason is that back in the day installing Cairo wasn't as easy as `pip install cairo-lang`, so an utility for installing was useful -- moreover with a new Cairo version every two weeks. I still believe there's value in setting up a whole environment from scratch, although we have `nile init` for that.

agree with everything else

[...]

PR #45, opened 2021-12-28
Add end-to-end tests for "compile" and "clean" commands

TimNooren:
See https://github.com/OpenZeppelin/nile/issues/41

Adds end-to-end tests for `nile compile` and `nile clean`. The latter was already pretty well covered by unit tests, so the test here is minimal.

Not sure about the best way to use markers here to in or exclude end-to-end tests. All feedback is welcome:)

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,8 @@ available task with `tox -av`.
 To run tests:
  * Install testing dependencies with `python -m pip install .[testing]`
  * Run all tests with `tox`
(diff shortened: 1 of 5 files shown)


fracek (approved):
LGTM.

fracek:
I think disabling the test for Python 3.10 is not an issue as (as you point out) it's not a version supported by Cairo.

Issue #35, opened 2021-12-09
Can't run nile node in macOS


kanishkatn:
Hi, 

I'm on macOS Big Sur, and there is some system process running at port 5000. Cause of which, `nile node` fails to run. 

Is there a way to specify the port in `nile node` ?

fracek:
At the moment it's not possible to change the host or port, we should do that.

Note: starknet devnet already supports this, we need to pass the host and port flags to it.

martriay:
Thanks for raising this! I'm on macOS but haven't stumbled upon this before. Would you be interested in opening a PR adding the feature @kanishkatn?

kanishkatn:
@martriay Yes, would love to!

milancermak:
Just FYI, there was a [post on Reddit](https://old.reddit.com/r/webdev/comments/qg8yt9/apple_took_over_port_5000_in_the_latest_macos/) about it some time ago. Seems the port squatting can be disabled via System preferences.

Yijia-Chen:
Second this issue. #46 seems to have a good solution already; can someone please merge?

martriay:
@Yijia-Chen hey there, bear in mind that merging it doesn't automatically makes it available since we'd have to also release a new version of nile. Until we release it, I suggest you clone the PR (or `main` when merged) and install it locally to use it now.

Yijia-Chen:
I installed the newest commit including #46 using `python -m pip install .` in the project directory, and `nile node` now has the flag `--port` which works on 5001. However, `nile deploy contract --alias contract_name` still has no `--port` option, and when I run it after `nile node`, it gives me the error that port 5000 is occupied. Am I doing something wrong here? Does the local `nile deploy` also use port 5000?

The detailed error message is as follows: `subprocess.CalledProcessError: Command '['starknet', 'deploy', '--contract', 'artifacts/contract.json', '--gateway_url=http://localhost:5000/']' returned non-zero exit status 1.` I am on macOS Monterey Version 12.0.1.

ftupas:
Hi @Yijia-Chen , normally when you run `nile node --port 5001` you'll have a `node.json` file with the corresponding gateway URL. No need to add the port when executing `nile deploy` as it will target whatever is saved in this file. However you do need to run it in the same directory when you did `nile node`, could you try again and see if it works on your end?

Yijia-Chen:
Okay now it works! I wasn't running them in the same directory. Thanks!

martriay:
[...]

PR #46, opened 2022-01-03
Add host and port options for "node" command

ftupas:
Hi there,

Kindly see #35. I've added an option to customize the host and node port when running `nile node` and an end-to-end test in `test_main.py`

Running the command `nile node --host localhost --port 5001` should yield something like this

image

Please let me know if this looks alright and any feedback is welcome, thanks! 🙂

### DIFF
--- a/src/nile/commands/deploy.py
+++ b/src/nile/commands/deploy.py
@@ -4,9 +4,7 @@ import re
 import subprocess
 
 from nile import deployments
-from nile.common import ABIS_DIRECTORY, BUILD_DIRECTORY
-
-GATEWAYS = {"localhost": "http://localhost:5000/"}
+from nile.common import ABIS_DIRECTORY, BUILD_DIRECTORY, GATEWAYS
 
 
 def deploy_command(contract_name, arguments, network, alias, overriding_path=None):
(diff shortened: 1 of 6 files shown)


fracek:
Looks good so far! Can you check why the tests are failing?

ftupas:
Thanks for the review, will look into it!

ftupas:
I've found the error,

ftupas:
It seems that the same error is thrown on `python 3.10`  like the end-to-end `test_compile` so I've added the same mark for failing.

image

This only covers the options for `node` command though. I realized I need more time to cover `deploy`, `call/invoke` but if this PR can be merged, then I'll try to solve them in a separate one if that's better. Please let me know 🙂 thank you!

ftupas:
Since [GATEWAY](https://github.com/OpenZeppelin/nile/blob/main/src/nile/common.py#L11) variable was static, `**deploy/call/invoke etc.**` commands wouldn't target the correct `gateway_url` if custom host or port was used so I've created a `node.json` file with default values `{"localhost": "http://localhost:5000/"}`. It gets overwritten with respective host and port whenever `nile deploy` is executed so the succeeding commands will work properly 🙂 Please let me know if this is okay, thank you!

fracek (approved):
Looks good to me. Thank you for the changes.

ftupas:
Happy to contribute, thanks for the review and approval 🙂

PR #52, opened 2022-01-18
Add setup.py to enable editable install

fracek:
At the moment, it's not possible to have an editable installation of nile (aka `pip install -e .[testing]`). I added a minimal `setup.py` that integrates with `setup.cfg` to enable this.


martriay:
Looks good, thanks!

Issue #29, opened 2021-11-09
Support scripting


martriay:
Expose a programatic API to execute `invoke`, `call` and `deploy` commands on any pre-configured network (local or mainnet by default), leveraging with the deployments management and eventually also [signers](https://github.com/OpenZeppelin/nile/issues/28).

fracek:
Should scripting be in Python (so nile exposes a Python API) or from a shell script?

martriay:
I think exposing python makes sense as of now, specially when you're already installing cairo and nile through pip.

l-henri:
Hello!
This looks like something I need. I'd like to write scripts similar to migrations in truffle, but in Nile.
How can I help?

martriay:
That's indeed the idea! Well I guess the first step would be to expose that functionality internally (`invoke`, `call`, `deploy`, etc) and then make the CLI commands depend on that "base" functionality. Once we have that, it should be automatically exportable and then we can modify the packaging system to include a programatic `nile` or `nile-lib` module :)_

l-henri:
Hello! Here are some things I would find very useful.
- Being able to send transactions through an account contract
- Being able to use an account contract both in Nile and argent X
- Pass hexadecimal values without having to convert to decimals
- Interact with new, as well as existing contracts
- Querying tx_status in the CLI to check wether a transaction passed or not

Super excited to be able to use this 😄

PR #55, opened 2022-01-23
Scripting API prototype

martriay:
Solves #29 

- [x] Replaces `print` with `logger` to allow silent execution (scripting API shouldn't write to `stdout`)
- [x] Renames `commands` to `core` to signal its now common lib for CLI commands and scripting API as well
- [x] Add `nile run` script to inject a stateful Nile's runtime environment object (managing `network` and whatnot for the user)

You can execute the `run(nre)` function of a script with `nile run path/to/script.py`

```python
def run(nre):
    address, abi = nre.deploy("contract", alias="my_contract")
    print(abi, address)
```

### DIFF
--- a/src/nile/common.py
+++ b/src/nile/common.py
@@ -1,6 +1,6 @@
 """nile common module."""
-import os
 import json
+import os
 
 CONTRACTS_DIRECTORY = "contracts"
 BUILD_DIRECTORY = "artifacts"
@@ -38,10 +38,3 @@ def get_all_contracts(ext=None):
         ]
 
     return files
-
-
-def logger(verbose):
-    def log(msg):
-        if verbose:
-            print(msg)
-    return log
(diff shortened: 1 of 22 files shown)


src/nile/common.py
```diff
@@ -38,3 +38,10 @@ def get_all_contracts(ext=None):
         ]
 
     return files
+
+
+def logger(verbose):
```
fracek, line 43: Should we use the Python's logger api? I think for now it's overkill.

martriay: heh I thought i was reinventing the wheel. you mean [this one](https://docs.python.org/3/library/logging.html) right?

martriay: removed this alltogether. might add it back later

src/nile/core/run.py
```diff
@@ -0,0 +1,14 @@
+"""Command to clean artifacts from workspace."""
```
fracek, line 1: Need to replace it with the description of the run command.

fracek (changes requested):
Looks great, just need to update one docstring and it's ready.

Tested locally and it works like a charm.

PR #56, opened 2022-01-25
Refactor account module and expose it through NRE

martriay:
* Refactor `account.py` and `signer.py`, created the class `Account` and redistributed responsibilities among modules
* Expose `Account` through NRE for scripting
* Remove `raw-execute` to simplify interface and docs


fracek (approved):
Everything looks good to me.

PR #60, opened 2022-02-14
feat(cli): improved --network option

milancermak:
This PR tries to improve the `--network` command line argument. It does the three following things:

1) The help message is better - it now enumerates the possible values for the argument (i.e. `localhost`, `goerli` or `mainnet`). The values are shown when using the `--help` arg or when passing in an invalid network name while using `--network`.

2) If takes the network value from the env var [`STARKNET_NETWORK`](https://starknet.io/docs/hello_starknet/account_setup.html#setting-up-the-network) which is what the `starknet` CLI tools use. If the env var is not set, it still defaults to `localhost`. If the env var is set, it can be overwritten by explicitly using the `--network` arg.

3) It normalizes the value of the `--network` argument so it's easier to switch toolchains (this one's highly subjective). Currently, there's no standardization in network naming. The Starknet CLI [uses](https://starknet.io/documentation/starknet-cli/) `alpha-goerli` and `alpha-mainnet`. Nile uses `goerli`, `mainnet` and `localhost`. Starknet.js [uses](https://github.com/0xs34n/starknet.js/blob/develop/src/provider/default.ts#L26) `mainnet-alpha` and `goerli-alpha`. Starknet.py [uses](https://github.com/software-mansion/starknet.py/blob/development/starknet_py/net/networks.py) `mainnet` and `testnet`. Might be that I'm the only one facing this issue, but sometimes I switch between projects/toolchains and it's a pain to try and remember what's the correct network name. In this PR I try to alleviate the pain a little by normalizing any string that has "goerli" or "testnet" in it to Nile's `goerli`. 

Please let me know what you think 😇

### DIFF
--- a/src/nile/cli.py
+++ b/src/nile/cli.py
@@ -18,6 +18,27 @@ from nile.core.version import version as version_command
 
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
 
+NETWORKS = ("localhost", "goerli", "mainnet")
+
+network_option = lambda f: click.option(  # noqa: E731
+    "--network",
+    envvar="STARKNET_NETWORK",
+    default="localhost",
+    help=f"Select network, one of {NETWORKS}",
+    callback=_validate_network,
+)(f)
+
+
+def _validate_network(_ctx, _param, value):
+    # normalize goerli
+    if "goerli" in value or "testnet" in value:
+        return "goerli"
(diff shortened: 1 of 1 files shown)


fracek (approved):
Looks good to me! I stumbled on the same issue myself so this is welcome.

martriay:
Awesome! :D

Issue #63, opened 2022-02-23
nile setup fails


ccarnino:
Hello team.
I am trying to setup a private key alias as described in the README. but I get an error.

I am on a working installation of Nile on macOS (I followed the macOS specific installations) and I am able to compile and deploy. I created the venv, installed cairo-nile with pip, inited the project. I also create a `.env` file in the root directory with the `YOLO=123` key.

```
> nile setup YOLO    

Traceback (most recent call last):
  File "/opt/homebrew/bin/nile", line 8, in <module>
    sys.exit(cli())
  File "/opt/homebrew/lib/python3.9/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/opt/homebrew/lib/python3.9/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/opt/homebrew/lib/python3.9/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/homebrew/lib/python3.9/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/homebrew/lib/python3.9/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/opt/homebrew/lib/python3.9/site-packages/nile/cli.py", line 84, in setup
    Account(signer, network)
  File "/opt/homebrew/lib/python3.9/site-packages/nile/core/account.py", line 20, in __init__
    self.signer = Signer(int(os.environ[signer]))
  File "/opt/homebrew/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9/os.py", line 679, in __getitem__
    raise KeyError(key) from None
KeyError: 'YOLO'
```

I have no idea what is going wrong

martriay:
Weird, I cannot reproduce it. Did you activate the virtualenv? With something like `source env/bin/activate`.

What directory are you running nile from? And where is your `.env`?

ccarnino:
So I have created a venv in the project directory, set the local python version to 3.7.12 (before it was 3.9 which seems not really compatible with nile).
Then I created a `.env` file inside the same root folder with `YOLO=1234`.

Then I run again:
```
❯ nile setup YOLO
🚀 Deploying Account
Got BadRequest
Traceback (most recent call last):
[...]

PR #74, opened 2022-03-08
Fix deployment issue with default gateway url

andrew-fleming:
This PR proposes to add `127.0.0.1` in favor of `localhost` as the default gateway url. The issue is explained below for posterity.

Deploying an account fails with a `BadRequest` error when, it appears, running MacOS. A suggested cause was that newer Macs use IPv6 which uses `::1/128` as its loopback network address as opposed to IPv4 which uses `127.0.0.1`. It seems, however, that this is not the entire cause. Turning off IPv6 and recloning the repo with a new venv still returned the `BadRequest` error. Further, it's confirmed that `localhost` was bound to `127.0.0.1` on my machine. I _was_ able to deploy an account with the default `localhost` while running Ubuntu. 

As proposed in #63, changing the default gateway url to `127.0.0.1` solves the deployment issue.

This PR also left the `localhost` nomenclature for aesthetics, but this can easily be changed if preferred.

### DIFF
--- a/src/nile/cli.py
+++ b/src/nile/cli.py
@@ -18,7 +18,7 @@ from nile.core.version import version as version_command
 
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
 
-NETWORKS = ("127.0.0.1", "goerli", "mainnet")
+NETWORKS = ("localhost", "goerli", "mainnet")
 
 
 def network_option(f):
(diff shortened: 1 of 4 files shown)


andrew-fleming:
To update: the `localhost` issue has to do with MacOS Monterey running AirPlay Receiver by default on port 5000. Users can simply turn it off; however, I think this change is still worth including.

martriay:
I don't think the port has anything to do with this, but reports are that this effectively fixes things for users, so let's go for it.

Issue #76, opened 2022-03-09
nile send failing with "TypeError: expected str, bytes or os.PathLike object, not int"


martinetlee:
Hello!

Using MacOS with Python 3.8.5, 
nile version 0.4.0

I have exported a environment variable `TEST_PRIV` to `123` and successfully run `nile setup TEST_PRIV` with the following output:
```
$ nile setup TEST_PRIV                     
🚀 Deploying Account
⏳ ️Deployment of Account successfully sent at 0x0636ab47cf615028c92a64e9aa9630d5bc93a335aa0ae174fbce2e3f22773cfc
🧾 Transaction hash: 0x40b2b24dd47c1d86079c9c8ecb1f30d556c6132e1247ef7063a82854f63a3fd
📦 Registering deployment as account-0 in localhost.deployments.txt
(base) 
```

now I wanted to interact with a deployed contract and hence running `nile send`
```
nile send TEST_PRIV token1 mint 0x0636ab47cf615028c92a64e9aa9630d5bc93a335aa0ae174fbce2e3f22773cfc 1000 
```

and I get the following error:

```
Traceback (most recent call last):
  File "/Users/redacted/opt/anaconda3/bin/nile", line 8, in <module>
    sys.exit(cli())
  File "/Users/redacted/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/Users/redacted/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/Users/redacted/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/redacted/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/redacted/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/Users/redacted/opt/anaconda3/lib/python3.8/site-packages/nile/cli.py", line 96, in send
    account.send(contract_name, method, params)
  File "/Users/redacted/opt/anaconda3/lib/python3.8/site-packages/nile/core/account.py", line 62, in send
    call_or_invoke(
[...]

PR #77, opened 2022-03-13
Fix signer <> account interaction

martriay:
This PR fixes the bug introduced in https://github.com/OpenZeppelin/nile/pull/65, solving https://github.com/OpenZeppelin/nile/issues/76.


serityfinance:
Hey @martriay @fracek just bumping this, this is currently blocking dev for us.

martriay:
@serityfinance I'm sorry about that! @andrew-fleming could you take a look at this PR?

In the mean time, I suggest you clone and install this PR to get it going.

src/nile/core/account.py
```diff
@@ -47,32 +46,32 @@ def deploy(self):
...
+            sender=self.address, calls=[[target_address, method, calldata]], nonce=nonce
         )
 
-        call_or_invoke(
+        params = []
+        params.append(str(len(call_array)))
+        params.extend([str(elem) for sublist in call_array for elem in sublist])
+        params.append(str(len(calldata)))
+        params.extend([str(param) for param in calldata])
+        params.append(str(nonce))
```
martriay, line 68: I don't like this part at all :(

andrew-fleming: Hmm I think this would look cleaner with a helper method to abstract away `str(len())` and `str()` + iteration. It's only 6 lines though, so maybe not worth it

src/nile/cli.py
```diff
@@ -97,7 +97,13 @@ def setup(signer, network):
 def send(signer, contract_name, method, params, network):
     """Invoke a contract's method through an Account. Same usage as nile invoke."""
     account = Account(signer, network)
-    account.send(contract_name, method, params)
+    print(
+        "Calling {} on {} with params: {}".format(
+            method, contract_name, [x for x in params]
+        )
+    )
+    out = account.send(contract_name, method, params)
```
andrew-fleming, line 105: Great addition

andrew-fleming (approved):
Looks great to me!

PR #79, opened 2022-03-14
Add account compilation check

andrew-fleming:
Cairo's 0.8.0 update requires that account contracts (with the external `__execute__` method) require an `--account_contract` flag to compile. This PR proposes to check for "Account" in the filename. In which case, the `--account_contract` flag will be included in the account compilation instruction. 

Note that this proposed change implies that all account contracts (again, with the exposed `__execute__` method) must have `Account` in its file name in order for Nile to correctly compile.

### DIFF
--- a/src/nile/core/compile.py
+++ b/src/nile/core/compile.py
@@ -58,6 +58,9 @@ def _compile_contract(path, directory=None):
         --output {BUILD_DIRECTORY}/{filename}.json \
         --abi {ABIS_DIRECTORY}/{filename}.json
     """
+    if "Account" in str(filename):
+        cmd = cmd + f"--account_contract"
+
     process = subprocess.Popen(cmd.split(), stdout=subprocess.PIPE)
     process.communicate()
     return process.returncode
--- a/tests/commands/test_compile.py
+++ b/tests/commands/test_compile.py
@@ -87,3 +87,28 @@ def test__compile_contract(mock_subprocess):
         stdout=mock_subprocess.PIPE,
     )
     mock_process.communicate.assert_called_once()
+
+
+def test__compile_account_contract(mock_subprocess):
+    contract_name_root = "Account"
+    path = f"path/to/{contract_name_root}.cairo"
+
+    mock_process = Mock()
+    mock_subprocess.Popen.return_value = mock_process
+
+    _compile_contract(path)
+
+    mock_subprocess.Popen.assert_called_once_with(
+        [
+            "starknet-compile",
+            path,
+            f"--cairo_path={CONTRACTS_DIRECTORY}",
+            "--output",
+            f"artifacts/{contract_name_root}.json",
+            "--abi",
+            f"artifacts/abis/{contract_name_root}.json",
+            f"--account_contract"
+        ],
+        stdout=mock_subprocess.PIPE,
+    )
+    mock_process.communicate.assert_called_once()


src/nile/core/compile.py
```diff
@@ -58,6 +58,9 @@ def _compile_contract(path, directory=None):
         --output {BUILD_DIRECTORY}/{filename}.json \
         --abi {ABIS_DIRECTORY}/{filename}.json
     """
+    if "Account" in str(filename):
+        cmd = cmd + "--account_contract"
```
martriay, line 62: what alternatives do we have? I think this could go wrong easily, like a non-account contract that says account (e.g. "AccountManager" could be a contract that manages accounts). Also we might want to expose this flag somehow with `nile compile`. For Nile's native accounts, we can simply pass the flag internally e.g. in `Account.py`'s constructor.

martriay (changes requested):
I think we should offer a more flexible API, like a flag to the `compile` command.

src/nile/core/compile.py
andrew-fleming: Great point! I was thinking of this as a temp fix so the Cairo Contracts weren't blocked. I'll work on integrating the flag in `nile compile`

src/nile/cli.py
```diff
@@ -144,7 +144,8 @@ def test(contracts):
 @cli.command()
 @click.argument("contracts", nargs=-1)
 @click.option("--directory")
-def compile(contracts, directory):
+@click.option("--account_contract", flag_value="--account_contract")
```
martriay, line 147: why is `flag_value="--account_contract"` necessary? instead of e.g. using `is_flag=True`

tests/resources/contracts/account_mock.cairo
```diff
@@ -0,0 +1,8 @@
+%lang starknet
+
+# The StarkNet compiler checks for an external "__execute__" method
+# if found, it expects an "--account_contract" flag during compilation
```
martriay, line 4: Is this true? I would have thought it's the other way around: only look for a `__execute__` method if the `--account_contract` flag is present.

martriay (approved):
Good work! I wonder how Cairo Contracts will behave now if we run a vanilla `nile compile` on the contracts folder, will it crash when trying to compile the Account?

src/nile/cli.py
andrew-fleming: I thought it might look cleaner, so the value that gets passed down to `core/compile` is either `False` or `--account_contract`. On the other hand, it might be easier to read in `core/compile` if the `--account_contract` string is hardcoded right there. Easy fix if preferred, what do you think?

andrew-fleming:
> I wonder how Cairo Contracts will behave now if we run a vanilla `nile compile` on the contracts folder, will it crash when trying to compile the Account?

It does crash, unfortunately. I haven't been able to come up with a good solution yet, but I think we need the CI to single out `Account` and `IAccount` somehow and compile with the flag (without running into the initial objection of checking for `Account` in the contract name). The good news is once they're compiled, the 0.8.0 update doesn't seem to require any further changes

tests/resources/contracts/account_mock.cairo
andrew-fleming: Yeah, it actually checks both ways

src/nile/cli.py
martriay: `--account_contract` is a CLI flag, while the `core/compile` module is a building block (ideally agnostic of external interfaces such as CLI, scripting API, or whatever). I wouldn't leak a CLI flag to the core module.

PR #80, opened 2022-03-15
Fix localhost, update readme

### LINKED ISSUE #78: Check python 3.9 issues
See https://github.com/OpenZeppelin/nile/issues/63#issuecomment-1054419137

andrew-fleming:
This PR proposes to further change `localhost` to `127.0.0.1`. The original change in #74 attempted to keep the `localhost` nomenclature. The `_validate_network` check rejects `127.0.0.1`, however, since it's looking for `localhost`, `goerli`, or `mainnet`.

This PR also updates the README to accurately show `127.0.0.1` instead of `localhost`. 

Finally, with these proposed changes, Python3.9 seems to be supported and resolves #78

Screen Shot 2022-03-15 at 3 14 43 PM
.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,15 +79,15 @@ nile deploy contract --alias my_contract
 
 🚀 Deploying contract
 🌕 artifacts/contract.json successfully deployed to 0x07ec10eb0758f7b1bc5aed0d5b4d30db0ab3c087eba85d60858be46c1a5e4680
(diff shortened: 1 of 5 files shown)


martriay:
I'm not sure about this change across the repo, particularly in the network files. What about taking advantage of `_validate_network` to normalize as we do with goearli?

PR #82, opened 2022-03-21
Add --disable-hint-validation flag to nile

LucasLvy:
This flag doesn't exist in nile so we it's impossible to compile smart-contracts with unwhitelisted hints such as print or other logging packages. Allowing them will significantly ease debugging

### DIFF
--- a/src/nile/core/compile.py
+++ b/src/nile/core/compile.py
@@ -11,13 +11,13 @@ from nile.common import (
 )
 
 
-def compile(contracts, directory=None, account_contract=False):
+def compile(
+    contracts, directory=None, account_contract=False, disable_hint_validation=False
+):
     """Compile cairo contracts to default output directory."""
     # to do: automatically support subdirectories
 
     contracts_directory = directory if directory else CONTRACTS_DIRECTORY
-
-    hint_validation = "--disable_hint_validation" if disable_hint_validation else ""
 
     if not os.path.exists(ABIS_DIRECTORY):
         logging.info(f"📁 Creating {ABIS_DIRECTORY} to store compilation artifacts")
@@ -32,7 +32,9 @@ def compile(contracts, directory=None, account_contract=False):
         all_contracts = get_all_contracts(directory=contracts_directory)
 
     results = [
-        _compile_contract(contract, contracts_directory, account_contract)
+        _compile_contract(
+            contract, contracts_directory, account_contract, disable_hint_validation
+        )
         for contract in all_contracts
     ]
     failed_contracts = [c for (c, r) in zip(all_contracts, results) if r != 0]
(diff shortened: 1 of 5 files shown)


src/nile/core/compile.py
```diff
@@ -11,12 +11,14 @@
...
 
 
-def compile(contracts, directory=None, account_contract=False):
+def compile(contracts, disable_hint_validation, directory=None, account_contract=False):
     """Compile cairo contracts to default output directory."""
     # to do: automatically support subdirectories
 
     contracts_directory = directory if directory else CONTRACTS_DIRECTORY
 
+    hint_validation = "--disable_hint_validation" if disable_hint_validation else ""
```
martriay, line 20: I suggest we pass the flag as a boolean to `_compile_contract` and let that function handle the boolean -- that's the only function that should be aware of compiler flags.

src/nile/core/compile.py
```diff
@@ -11,12 +11,14 @@
 )
 
 
-def compile(contracts, directory=None, account_contract=False):
+def compile(contracts, disable_hint_validation, directory=None, account_contract=False):
```
martriay, line 14:
```suggestion
def compile(contracts, directory=None, account_contract=False, disable_hint_validation=False):
```
I suggest moving this option to the end of the argument list since I expect the other ones to be used more often. Also, having it first forces users to pass a value even if they don't care about the option.

martriay (changes requested):
Looking good, thanks for this! Can you also add a test for the feature, as well as a small mention in the README so it's documented? Thanks!

src/nile/core/compile.py
LucasLvy: Sure lets move it to the end though as it is a flag if you don't add it it'll automatically be set to `False`

martriay: Not necessarily, Nile can be used programmatically not just through CLI

martriay: see https://github.com/OpenZeppelin/nile/#run

tests/commands/test_compile.py
```diff
@@ -112,3 +112,28 @@ def test__compile_account_contract(mock_subprocess):
...
+            f"--cairo_path={CONTRACTS_DIRECTORY}",
+            "--output",
+            f"artifacts/{contract_name_root}.json",
+            "--abi",
+            f"artifacts/abis/{contract_name_root}.json",
+            "--disable_hint_validation",
+        ],
+        stdout=mock_subprocess.PIPE,
+    )
+    mock_process.communicate.assert_called_once()
```
martriay, line 139: is this enough to confirm it ran successfully? so it makes sense to have an example contract

martriay (approved):
Looks good :)

Please run the linter as [specified in the readme](https://github.com/OpenZeppelin/nile#hacking-on-nile) so we can merge it

tests/commands/test_compile.py
LucasLvy: I'm not used to testing with `Mock` so i don't know. I used the same test as for the account flag. To me it would make sense to assert that the command exited with 0 and to try compiling a contract with unwhitelisted hints without adding the flag and to assert that it failed but once more i'm not really familiar with the mock class

PR #85, opened 2022-03-23
fix import

### LINKED ISSUE #84: Import error during Nile installation
After a fresh install of `cairo-nile`, attempts to `nile init` will return a `ModuleNotFoundError: No module named 'starkware'` error. Here's the traceback:

```
(env) andrewfleming@Andrews-MacBook-Pro nile % nile init
  File "/Users/andrewfleming/Documents/starknet/nile-delete-me/env/bin/nile", line 6, in <module>
    from nile.cli import cli
  File "/Users/andrewfleming/Documents/starknet/nile-delete-me/env/lib/python3.7/site-packages/nile/cli.py", line 7, in <module>
    from nile.core.account import Account
  File "/Users/andrewfleming/Documents/starknet/nile-delete-me/env/lib/python3.7/site-packages/nile/core/account.py", line 9, in <module>
    from nile.signer import Signer
  File "/Users/andrewfleming/Documents/starknet/nile-delete-me/env/lib/python3.7/site-packages/nile/signer.py", line 3, in <module>
    from starkware.cairo.common.hash_state import compute_hash_on_elements
ModuleNotFoundError: No module named 'starkware'
```

TL;DR:
- nile imports from `nile/cli`
- `nile/cli` imports from `nile/core/account`
- `nile/core/account` imports from `nile/signer`
- `nile/signer` imports from starkware (which is not installed until after initialization)

Wrapping `nile/core/account`'s import of `nile/signer` should rectify this issue.

andrew-fleming:
Resolves #84.

### DIFF
--- a/src/nile/core/account.py
+++ b/src/nile/core/account.py
@@ -6,7 +6,10 @@ from dotenv import load_dotenv
 from nile import accounts, deployments
 from nile.core.call_or_invoke import call_or_invoke
 from nile.core.deploy import deploy
-from nile.signer import Signer
+try:
+    from nile.signer import Signer
+except ImportError:
+    pass
 
 load_dotenv()
 


martriay:
This prevents the error, but is it right? Should the user be able to continue even if `starknet` is not installed?

andrew-fleming:
I was thinking in terms of following the setup guide. All `nile` commands are blocked if `starknet` isn't already installed, but it gets installed after `nile init`. I suppose we could just amend the guide and let users know that they should install `cairo-lang` first—that was my working solution prior to this PR. What do you think?

PR #86, opened 2022-03-25
Plugin support

### LINKED ISSUE #53: Support Nile plugins
Add a plugin system to extend Nile's functionality

franalgaba:
Fixes #53 

# Plugin support for Nile

In this PR I present a methodology for extending CLI functionality for Nile in a Pythonic manner to ease Nile extensibility with plugins. 

## How is this implemented? 

This implementation takes advantage of the native extensibility features of click. Using click and leveraging the Python entrypoints we have a simple manner of handling extension natively on Python environments through dependencies. The plugin implementation on Nile looks for specific Python entrypoints constraints for adding commands.

In order for this implementation to be functional, it is needed by the plugin developer to follow some development guidelines defined in [this simple plugin example](https://github.com/franalgaba/nile-plugin-example) I've implemented extending Nile for a dummy greet extension. In a brief explanation the guidelines are as follows:

1. Define a Python module that implement a click command or group:

```python
# First, import click dependency
import click

# Decorate the method that will be the command name with `click.command` 
@click.command()
# You can define custom parameters as defined in `click`: https://click.palletsprojects.com/en/7.x/options/
def my_command():
    # Help message to show with the command
    """
    Subcommand plugin that does something.
    """
    # Done! Now implement your custom functionality in the command
    click.echo("I'm a plugin overiding a command!")
```

2. Define the plugin entrypoint. In this case using Poetry features in the `pyproject.toml` file:

```
# We need to specify that click commands are Poetry entrypoints of type `nile_plugins`. Do not modify this
[tool.poetry.plugins."nile_plugins"]
# Here you specify you command name and location <command_name> = <package_method_location>
"greet" = "nile_greet.main.greet"
```

3. Done!

How to decide if I want to use a plugin or not? Just install / uninstall the plugin dependency :)

Cheers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,9 +206,9 @@ Nile has the possibility of extending its CLI and `NileRuntimeEnvironment` funct
 
 ### How it works
 
-This implementation takes advantage of the native extensibility features of click. Using click and leveraging the Python entrypoints we have a simple manner of handling extension natively on Python environments through dependencies. The plugin implementation on Nile looks for specific Python entrypoints constraints for adding commands.
+This implementation takes advantage of the native extensibility features of [click](https://click.palletsprojects.com/). Using click and leveraging the Python [entrypoints](https://packaging.python.org/en/latest/specifications/entry-points/) we have a simple manner of handling extension natively on Python environments through dependencies. The plugin implementation on Nile looks for specific Python entrypoints constraints for adding commands.
 
-In order for this implementation to be functional, it is needed by the plugin developer to follow some development guidelines defined in this simple plugin example I've implemented extending Nile for a dummy greet extension. In a brief explanation the guidelines are as follows:
+In order for this implementation to be functional, it is needed by the plugin developer to follow some development guidelines defined in this simple plugin example extending Nile for a dummy greet extension. In a brief explanation the guidelines are as follows:
 
 1. Define a Python module that implement a click command or group:
 
@@ -240,6 +240,8 @@ def my_command():
 3. Done!
 
 How to decide if I want to use a plugin or not? Just install / uninstall the plugin dependency from your project :smile:
+
+Finally, after the desired plugin is installed, it will also be automatically available through the `nre`. The plugin developer should be aware of this and design the interface accordingly.
 
 ## Hacking on Nile
 
--- a/README.md
+++ b/README.md
@@ -200,6 +200,47 @@ Print out the Nile version
 nile version
 ```
 
+## Extending Nile with plugins
+
+Nile has the possibility of extending its CLI and `NileRuntimeEnvironment` functionalities through plugins. For developing plugins for Nile fork [this plugin example](https://github.com/franalgaba/nile-plugin-example) boilerplate and implement your desired functionality with the provided instructions.
+
+### How it works
+
+This implementation takes advantage of the native extensibility features of click. Using click and leveraging the Python entrypoints we have a simple manner of handling extension natively on Python environments through dependencies. The plugin implementation on Nile looks for specific Python entrypoints constraints for adding commands.
+
+In order for this implementation to be functional, it is needed by the plugin developer to follow some development guidelines defined in this simple plugin example I've implemented extending Nile for a dummy greet extension. In a brief explanation the guidelines are as follows:
+
+1. Define a Python module that implement a click command or group:
+
+```python
+# First, import click dependency
+import click
+
+# Decorate the method that will be the command name with `click.command` 
+@click.command()
+# You can define custom parameters as defined in `click`: https://click.palletsprojects.com/en/7.x/options/
+def my_command():
+    # Help message to show with the command
+    """
+    Subcommand plugin that does something.
+    """
+    # Done! Now implement your custom functionality in the command
+    click.echo("I'm a plugin overiding a command!")
+```
+
+2. Define the plugin entrypoint. In this case using Poetry features in the pyproject.toml file:
+
+```
+# We need to specify that click commands are Poetry entrypoints of type `nile_plugins`. Do not modify this
+[tool.poetry.plugins."nile_plugins"]
+# Here you specify you command name and location <command_name> = <package_method_location>
+"greet" = "nile_greet.main.greet"
+```
+
+3. Done!
+
+How to decide if I want to use a plugin or not? Just install / uninstall the plugin dependency from your project :smile:
+
 ## Hacking on Nile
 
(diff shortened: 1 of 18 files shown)


martriay:
This is really cool, thanks! But I think we're still missing making plugins available through Nile's programmatic API

franalgaba:
> This is really cool, thanks! But I think we're still missing making plugins available through Nile's programmatic API

Thanks Martín! Could you give more details about the missing functionality? Do you mean importing the plugins directly from Nile's module? Will work on that if it's what is needed.

martriay:
I mean exposing the plugin's functionality through [NileRuntimeEnvironment](https://github.com/OpenZeppelin/nile/#run)

franalgaba:
> I mean exposing the plugin's functionality through [NileRuntimeEnvironment](https://github.com/OpenZeppelin/nile/#run)

Just added the implementation for installed plugin's functionality to be exposed in the NRE :)

Let me know what you think.

fracek:
I went through the PR and I really like it a lot. I only have two concerns:

* The command added to the `nre` object should be a method and not a static function. 
* At the moment the core commands and `nre` methods are two separate functions.  Is there any downside in having commands-only extensions? How does it work when the command takes arguments?

src/nile/nre.py
```diff
@@ -4,11 +4,15 @@
 from nile.core.call_or_invoke import call_or_invoke
 from nile.core.compile import compile
 from nile.core.deploy import deploy
+from nile.core.utils import _get_installed_plugins, skip_click_exit
 
 
 class NileRuntimeEnvironment:
     """The NileRuntimeEnvironment exposes Nile functionality when running a script."""
 
+    for name, object in _get_installed_plugins().items():
```
fracek, line 13: I believe that if we pass the plugins dict in a `__init__` function then it's possible to also add tests for this. It requires a change in `run` to load the plugins and then instantiate `nre`.

franalgaba:
> I went through the PR and I really like it a lot. I only have two concerns:
> 
> * The command added to the `nre` object should be a method and not a static function.
> * At the moment the core commands and `nre` methods are two separate functions.  Is there any downside in having commands-only extensions? How does it work when the command takes arguments?

Hi @fracek . Thank you for the comments. The questions you asked have a reasoning:

* The command added to the `nre` object should be a method and not a static function.

The `nre` object is a class. If you don't define the plugin command as a `staticmethod` you will be forced to add a `self` parameter in the command that won't be used by itself and I think it is a bit weird from the plugin development side. If you just don't add the `self` parameter to the command and don't define it as a `staticmethod` you will get the following error: `TypeError: <command> takes 0 positional arguments but 1 was given` which is pretty logical because the `self` object is being passed in but is not expected. In my opinion, that is the most logical solution. Let me know if you know any better implementation :)

* At the moment the core commands and nre methods are two separate functions. Is there any downside in having commands-only extensions? How does it work when the command takes arguments?

The plugin extensions are not limited by commands. It can be everything that is supported by `click` natively. In addition to commands there can be `click.group` as plugins as well for example. Regarding the arguments, when a click command has options defined they are given to the method as keyword arguments so there are no implications as far as I know. You cand find more info [here](https://click.palletsprojects.com/en/8.0.x/commands/)

fracek:
I really like it now! I think the only open issue now is how to test it. Unfortunately I also always had issues testing entry-point based code.

franalgaba:
> I really like it now! I think the only open issue now is how to test it. Unfortunately I also always had issues testing entry-point based code.

Workin on it :)

franalgaba:
- Added tests for the `nile.core.utils` module for validation plugin loading.
- Added tests for the `nile.nre` module for plugin availability in NRE.
- Added coverage check support in tox.ini.

I think there isn't anything else missing. Let me know if need something else :)

martriay:
I cannot reproduce this on my machine :/
`nile node` and `nile node --host localhost --port 5001` work just fine, so do the tests. I'm using `cairo-lang@0.8.0`, `starknet-devnet@0.1.19` and `nile@0.5.2`

davidism:
The `WERKZEUG_SERVER_FD` indicates that something you're using is messing with internal variables, which puts the environment in an unexpected state when Werkzeug tries to start. In gdbui's case, they were setting `WERKZEUG_RUN_MAIN` manually, instead of disabling the reloader by passing `use_reloader=False`.

In your traceback, it looks like the problem comes from starknet-devnet. In their v0.1.19 release, they are doing the same thing https://github.com/Shard-Labs/starknet-devnet/blob/v0.1.19/starknet_devnet/server.py#L308. In their development branch they have switched to meinheld instead of the Werkzeug dev server, so the issue will no longer be present once they make a new release.

src/nile/core/utils.py
```diff
@@ -0,0 +1,57 @@
...
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except SystemExit:
+            pass
+
+    return wrapper
+
+
+def _get_installed_plugins() -> Dict:
```
martriay, line 28: why the `_` if it's meant to be exported/used by other modules?

README.md
```diff
@@ -200,6 +200,47 @@ Print out the Nile version
...
 
+## Extending Nile with plugins
+
+Nile has the possibility of extending its CLI and `NileRuntimeEnvironment` functionalities through plugins. For developing plugins for Nile fork [this plugin example](https://github.com/franalgaba/nile-plugin-example) boilerplate and implement your desired functionality with the provided instructions.
+
+### How it works
+
+This implementation takes advantage of the native extensibility features of click. Using click and leveraging the Python entrypoints we have a simple manner of handling extension natively on Python environments through dependencies. The plugin implementation on Nile looks for specific Python entrypoints constraints for adding commands.
+
+In order for this implementation to be functional, it is needed by the plugin developer to follow some development guidelines defined in this simple plugin example I've implemented extending Nile for a dummy greet extension. In a brief explanation the guidelines are as follows:
```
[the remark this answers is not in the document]
martriay: Also should we mention that every command gets automatically added to `nre` and that interfaces should be designed accordignly?

martriay (commented):
Love it! I left a few comments. I think we should open a new issue to design hook-kind of plugins as [envisioned by @fracek](https://github.com/fracek/nile/commit/b250500e263ab5a4faa920bc223a8754cf98a676) in  #53

martriay:
Thank you all! Impressive first contribution @franalgaba  👏

PR #90, opened 2022-03-28
Add locate-error command

### LINKED ISSUE #50: Debug contracts with Nile
It would be nice to debug contracts. I wrote a [Cairo debugger](https://github.com/fracek/cairo-dap) that uses the debugger adapter protocol to communicate with VS Code, Emacs, and Vim. The package doesn't work with recent versions of Cairo or StarkNet (that was before StarkNet was announced).

How should the debugger be started? I was thinking it makes sense to pass the contract, a method, and its arguments.

andrew-fleming:
Fixes #50. This PR supercedes #70 and proposes to simplify the `RETRY_AFTER_PERIOD` unit/value mapping to a simple `RETRY_AFTER_SECONDS` variable.


trevis-dev:
Hello @andrew-fleming , glad to see the feature consolidating! 

I understand you went for the `RETRY_AFTER_SECONDS` option, but the variable isn't (yet ?) named like that.

If I may, I do think it'd be better than the current `_TIME` and just wanted to point it out in case it was simply forgotten/overlooked.

tests/test_cli.py
```diff
@@ -141,3 +143,27 @@ def test_node(args, expected):
...
+    [
+        ([MOCK_HASH, "--network", "goerli"]),
+        ([MOCK_HASH, "--network", "mainnet", "--contracts_file", "example.txt"]),
+    ],
+)
+@patch("nile.utils.debug.subprocess")
+def test_locate_error(mock_subprocess, args):
+    # locate-error will hang without patch
+    mock_json = json.dumps({"tx_status": "ACCEPTED"})
+    mock_subprocess.check_output.return_value = mock_json
```
martriay, line 159:
```suggestion
    mock_subprocess.check_output.return_value = json.dumps({"tx_status": "ACCEPTED"})
```
any reason not to?

martriay, line 159:
```suggestion
    mock_subprocess.check_output.return_value = json.dumps({"tx_status": "ACCEPTED"})
```
any reason not to?

martriay (approved):
Looks very good overall! Great work @andrew-fleming and @trevis-dev 🚀 

Two things come to mind:

1. given the issue and the function are called `debug`, shouldn't we call the command the same way? or does `nile debug` convey a different idea to you?
2. I think this await/retry strategy could work well for an `observe` or similar command that checks the state of a sent tx and updates with the latest, until it has a final value. This could even be a flag for `send`/`invoke`.

tests/test_cli.py
andrew-fleming: Nope, good call!

andrew-fleming:
@martriay hmmm those are great points to consider. 
1. I think there's a nice simplicity to `nile debug` and it keeps up with the single word commands. I don't have a very good reason not to opt for `nile debug` outside of preference. What do you think @trevis-dev?
2. I agree! It'd be a super convenient feature to just add a flag in `send` or `invoke` and not have to manually check the tx status thereafter

trevis-dev:
@martriay , @andrew-fleming 

Great ideas, indeed!

1. I think `debug` is reasonable, especially considering that different features could still be implemented in the future if something other than a tx hash is passed as main argument.

2. I completely agree as well. In fact I have been using this command to watch any transaction's status, also as a user and not a developer of the related contracts.

martriay:
Awesome! Let's rename it then, and open an issue for the other command/flag :)

andrew-fleming:
> I have been using this command to watch any transaction's status, also as a user and not a developer of the related contracts.

@trevis-dev Haha so have I!

> Awesome! Let's rename it then, and open an issue for the other command/flag :)

@martriay Done and done!

martriay (approved):

martriay:
Amazing work guys!

PR #92, opened 2022-04-01
Fix werkzeug version

martriay:
Fix [upstream dependency error](https://github.com/pallets/werkzeug/issues/2364) by downgrading it. Thanks to @franalgaba for [looking into it](https://github.com/OpenZeppelin/nile/pull/86#issuecomment-1084985907).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -12,6 +12,7 @@ passenv =
 extras =
     testing
 deps =
+    werkzeug==2.0.0
     starknet-devnet
 commands =
     pytest {posargs}


davidism:
I think you linked the wrong issue that this pin addresses, but regardless the linked issue regarding `cache_control.s_maxage` is now fixed in Werkzeug 2.1.1. I'll comment about the env var issue on the linked nile issue.

davidism:
I think you linked the wrong issue that this pin addresses, but regardless the linked issue regarding `cache_control.s_maxage` is now fixed in Werkzeug 2.1.1. I'll comment about the env var issue on the linked nile issue.

martriay:
Thanks! And you're right

PR #112, opened 2022-05-01
Automate account flag during compilation

### LINKED ISSUE #102: Automating the --account_flag during compilation
When running `nile compile`, there should be a standardized way to let the compiler know that a contract is an account and thus automatically insert the `--account_contract` flag. My initial strategy was to simply look for `Account` in the contract name, but after the discussion in #79, this appears to be too restrictive. A better approach might be to trigger the flag when `Account` is the suffix i.e. `Account.cairo` and `<something>Account.cairo`.

andrew-fleming:
This PR proposes to automatically include the `--account_contract` flag during compilation if the contract name ends with `Account`. If this is accepted, it:
- sets a precedent that account contracts should suffix their name with `Account`
  - includes the common Account.cairo
- solves the issue of compiling entire libraries such as [OpenZeppelin's Contracts for Cairo](https://github.com/OpenZeppelin/cairo-contracts)
- will not trigger for contract names such as AccountManager.cairo
- can still be triggered for uncommon contract names manually
- sets the sole restriction that non-account contracts should not suffix their name with `Account`

This resolves #102.

---

I think it'd be a good idea to add a directive (`%`) for account contracts as mentioned [here](https://github.com/OpenZeppelin/nile/issues/102#issuecomment-1104274490) by @JulissaDantes. This will probably have to be done in the [StarkNet compiler.](https://github.com/starkware-libs/cairo-lang/tree/64a7f6aed9757d3d8d6c28bd972df73272b0cb0a/src/starkware/starknet/compiler)

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,10 +64,10 @@ nile compile contracts/MyContract.cairo # compiles single contract
 nile compile contracts/MyContract.cairo --disable-hint-validation # compiles single contract with unwhitelisted hints
 ```
 
-As of cairo-lang v0.8.0, account contracts (contracts with the `__execute__` method) must be compiled with the `--account_contract` flag.
+As of cairo-lang v0.8.0, account contracts (contracts with the `__execute__` method) must be compiled with the `--account_contract` flag. Nile automatically inserts the flag if the contract's name ends with `Account` i.e. Account.cairo, EthAccount.cairo. Otherwise, the flag must be included by the user.
 
 ```sh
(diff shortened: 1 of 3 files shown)


archseer:
Is this case sensitive? I've been using `account.cairo`/snake case names rather than camel case.

martriay:
feel free to open a PR supporting it

andrew-fleming:
Hey, @archseer! Yes, this is case sensitive for the automated flag, but you can still manually set it. Adding snake cased `account` might be a good idea :)

PR #114, opened 2022-05-04
Fix alias in debug

### LINKED ISSUE #103: nile debug contract identification fails due to aliases
https://github.com/OpenZeppelin/nile/blob/915c7f47b5f9fdde80598435d538f1aa81d42526/src/nile/utils/debug.py#L68

If your deployments feature an alias, you have ADDRESS:ABI:ALIAS,
which causes the line above to fail.

Suggested solution:
`line_address, abi, *_ = line.split(":")`

andrew-fleming:
This PR fixes the alias bug in `debug` and proposes to break some of the functionality away from `debug` into a separate internal function (for easier testing). This resolves #103.

### DIFF
--- a/tests/commands/test_debug.py
+++ b/tests/commands/test_debug.py
@@ -48,9 +48,9 @@ def test__locate_error_lines_with_abis_with_and_without_alias(
     with patch("nile.utils.debug.open") as mock_open:
         mock_open.return_value.__enter__.return_value = file
         return_array = _locate_error_lines_with_abis(MOCK_FILE, address_set, mock_path)
-        # The contract alias should be omitted in the output
+        # Values should be pushed into the array (and not return an error)
+        # regardless if a contract alias is present or not
         assert return_array == [f"{DEBUG_ADDRESS}:{ABI_PATH}"]
-        assert ALIAS not in return_array[0]
 
 
 @patch("nile.utils.debug._abi_to_build_path", return_value=ABI_PATH)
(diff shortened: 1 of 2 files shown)


martriay (approved):
Looks very good, thanks!

tests/commands/test_debug.py
```diff
@@ -30,6 +34,36 @@ def test__abi_to_build_path():
...
+@patch("nile.utils.debug._abi_to_build_path", return_value=ABI_PATH)
+def test__locate_error_lines_with_abis_with_and_without_alias(
+    mock_path, file, address_set
+):
+    with patch("nile.utils.debug.open") as mock_open:
+        mock_open.return_value.__enter__.return_value = file
+        return_array = _locate_error_lines_with_abis(MOCK_FILE, address_set, mock_path)
+        # The contract alias should be omitted in the output
+        assert return_array == [f"{DEBUG_ADDRESS}:{ABI_PATH}"]
+        assert ALIAS not in return_array[0]
```
martriay, line 53: I'm not sure it's that important that the alias is dropped, but that it works when there's one

martriay, line 53: I'm not sure it's that important that the alias is dropped, but that it works when there's one

andrew-fleming: Good question. The latter is the priority, I'll remove the `ALIAS` assertion.

PR #119, opened 2022-05-16
Add get-accounts command

### LINKED ISSUE #87: Get accounts
Add functionality to manage/retrieve available accounts on Nile, to use them programmatically.

andrew-fleming:
This PR adds the `get-accounts` command to Nile's core functionality. In order for the NRE to have access to the private key to sign transactions, this PR proposes to include the private key alias in account registration—thus, storing it in the `<networks>.accounts.json` module. 

The `get-accounts` feature is meant to be a plugin; however, this PR shows the basic logic as if it were part of the core functionality. The changes necessary to make this a plugin should be fairly simple and straightforward—assuming the logic is acceptable.

Resolves #87.


JulissaDantes (approved):
Looks good to me

README.md
```diff
@@ -346,6 +346,41 @@ CONTRACT_ADDRESS2:PATH_TO_COMPILED_CONTRACT2.json
 ...
 ```
 
+### `get-accounts`
+
+Retrieves a list of ready-to-use accounts which allows for easy scripting integration. Before using `get-accounts`:
+
+1. store private keys in a `.env`
```
martriay, line 353:
```suggestion
1. store private keys in a `.env`

    ```
    PRIVATE_KEY_ALIAS_1=286426666527820764590699050992975838532
    PRIVATE_KEY_ALIAS_2=263637040172279991633704324379452721903
    PRIVATE_KEY_ALIAS_3=325047780196174231475632140485641889884
    ```

```

src/nile/utils/get_accounts.py
```diff
@@ -0,0 +1,41 @@
+"""Retrieve and manage deployed accounts."""
+import json
+import logging
+
+from nile.accounts import current_index
+from nile.core.account import Account
+
+
+def _check_and_return_account(signer, pubkey, network):
```
martriay, line 9: what do you think about moving this to the bottom?

martriay (approved):
Code looks very good! My sole concern is that this doesn't integrate with the devnet and there's two sets of accounts: devnet's and nile's. The former are pre-funded and the latter are usually user-deployed. I wonder if nile should pick up the devnet's when it's local or not.

At least to avoid/help out confusion, we could change the `get-accounts` stdout messaging so it says "Total registered accounts" and somehow convey that they're the ones deployed through `nile setup`, maybe just a link to the docs?

tests/commands/test_get_accounts.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@pytest.mark.parametrize(
+    "private_keys, public_keys",
+    [
+        ([ALIASES[0], PUBKEYS[1]]),
+        ([ALIASES[1], PUBKEYS[0]]),
+    ],
+)
+def test__check_and_return_account_with_mismatching_keys(private_keys, public_keys):
```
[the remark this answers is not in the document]
andrew-fleming: I chose to double underscore because the func begins with an underscore (`_check_and_return_account`). I'll change it to a single, it looks a bit funky now

andrew-fleming:
> I wonder if nile should pick up the devnet's when it's local or not.

I think this is a good idea. If we integrate something like this, it should probably be included in #137 or at least loosely coordinated.

> At least to avoid/help out confusion, we could change the `get-accounts` stdout message so it says "Total registered accounts" and somehow convey that they're the ones deployed through `nile setup`, maybe just a link to the docs?

Yeah, that sounds good. I'll update accordingly.

tests/commands/test_get_accounts.py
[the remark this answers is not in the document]
martriay: nah, it's good

martriay:
> I think this is a good idea. If we integrate something like this, it should probably be included in #137 or at least loosely coordinated.


Sure that issue tracks it better. We shouldn't take long between releasing this and #137, it can be confusing/annoying not be able to use devnet's.

andrew-fleming:
> > I think this is a good idea. If we integrate something like this, it should probably be included in #137 or at least loosely coordinated.
> 
> Sure that issue tracks it better. We shouldn't take long between releasing this and #137, it can be confusing/annoying not be able to use devnet's.

Yes, agreed

tests/test_nre.py
```diff
@@ -31,3 +30,31 @@ def dummy_params(a, b):
...
         assert 3 == nre_result
+
+
+def test_nre_get_accounts(capsys):
+    nre = NileRuntimeEnvironment()
+    with patch("nile.nre.get_accounts", return_value=(123)):
+        res = nre.get_accounts()
+        captured = capsys.readouterr()
+
+        assert res == (123)
```
martriay, line 41: isn't this always true because of the function being patched above? wouldn't that be independent from the implementation under test?

martriay (approved):
Looks good! Left just a question

tests/test_nre.py
andrew-fleming: Correct! Testing for the actual implementation is in [`tests/commands/test_get_accounts.py`](https://github.com/andrew-fleming/nile/blob/get-accounts/tests/commands/test_get_accounts.py). I added a conditional in the NRE (which is what this was testing), but good call pointing this out. I see a better way to simplify this

martriay (approved):

PR #122, opened 2022-05-22
add `status` command and chaining.

### LINKED ISSUE #98: Add `status` command
Add `observe` command (or some other name) that fetches the status of a transaction using an await/retry strategy. Ideally, this should be an optional flag for `send`, `invoke`, and `deploy`. See #90 for discussion and its implementation in `nile debug`.

trevis-dev:
Fixes #98. This PR implements the `status` command to fetch the status of a StarkNet transaction and the ability to wait for its settlement before initiating another transaction.

See discussion in #98. Follow up of #70, #90.

The PR splits features currently available in `nile debug` and implements them in other functions to be used in different cases:
  * From any transaction hash, query its result. `nile status <transaction_hash>`.
  * If the transaction may not be settled at calling time and the user wishes to wait (and get that status) before continuing, the `--track` flag can be added. `status` will then retry on pending transactions until they are REJECTED or ACCEPTED IN L2.°
  * If users would like to debug the transaction message in case of rejection, use the `--debug` flag.  
  Note that `--debug` implies `--track`. The `debug`command keeps the same functionality and is effectively an alias for `status --debug`. 

Furthermore, `nile invoke`, `nile deploy`, `nile declare` `Account.send`, and `Account.deploy` can now also use `--track` and `--debug`. 
In that case, the process (terminal or within `run`) will not return before the transaction is settled (helping the fight against "nonce"). 

`status`, as well as `nile invoke` and `account.send` when these flags are used, returns a `TransactionStatus`, named tuple holding the transaction hash, the transaction status, and the eventual error message.
Currently, `deploy`s return value is kept unchanged. 

° An additional flag `--wait-for-l1` was also on my mind but I didn't think it was necessary for now.


EvolveArt:
Exactly what's needed so nile scripts can reach wider adoption. They are very powerful but still truly underused because lack such features. Right now, you can't execute multiple transactions in a row on a script. This is truly needed and huge thanks to @trevis-dev for this PR 🤝

andrew-fleming:
@trevis-dev this looks awesome! Thank you for getting this done, sir. Just a friendly heads up: we're a little busy with the contracts library right now, but we'll get to this as soon as we can :) I'm excited to check this out 🚀

martriay:
@trevis-dev could you fix the conflicts so we can resume work on this? :)

trevis-dev:
Hey guys! Can I do something else for this while it's still fresh in memory?

trevis-dev:
Could I just ask you please to *add* to these commits if another update is required before merging, when you are ready?

andrew-fleming:
Hey @trevis-dev! I'm so sorry for the delay. I tested this a few times and got stuck reconciling updates (and then we introduced new changes). We'll certainly add to these commits. Thank you for making this PR 🙏 I know you put a lot of work into it

andrew-fleming:
I think it's worth mentioning that status queries won't work with the NRE until #215 is resolved.

src/nile/cli.py
```diff
@@ -89,26 +111,42 @@ def run(path, network):
...
 @cli.command()
 @click.argument("signer", nargs=1)
 @network_option
-def setup(signer, network):
+@track_option
+@debug_option
+def setup(signer, network, track=False, debug=False):
     """Set up an Account contract."""
-    Account(signer, network)
+    Account(signer, network=network, track=track, debug=debug)
```
ericnordelo, line 149: A small suggestion to avoid to much verbosity when is possible:
```suggestion
    Account(signer, network, track=track, debug=debug)
```

src/nile/common.py
```diff
@@ -122,3 +117,18 @@ def is_string(param):
 def is_alias(param):
     """Identiy param as alias (instead of address)."""
     return is_string(param)
+
+
+def get_network_parameter(network):
```
ericnordelo, line 122: I think is a little confusing that a "getter" writes to the env (for code readability). I will suggest changing the name of the function to something like `get_network_param_or_set_env`.

src/nile/core/call_or_invoke.py
```diff
@@ -76,3 +79,15 @@ def call_or_invoke(
             )
 
         return ""
+
+    if type != "call" and output:
+        logging.info(output)
+        transaction_hash = _get_transaction_hash(output)
+        return status(normalize_number(transaction_hash), network, track, debug)
```
ericnordelo, line 86: Nice improvement!

ericnordelo (commented):
Looks good @andrew-fleming! I left a couple of small suggestions. And I want to add that `watch` sounds better to me than `track` as the option for connecting the terminal waiting for status changes, just for consideration.

src/nile/cli.py
```diff
@@ -57,6 +57,28 @@ def _validate_network(_ctx, _param, value):
     raise click.BadParameter(f"'{value}'. Use one of {NETWORKS}")
 
 
+def track_option(f):
```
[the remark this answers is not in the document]
ericnordelo: We could have something like:
```
...
@click.option("--track", "status_type", flag_value="track")
@click.option("--debug", "status_type", flag_value="debug")
...
```

Handling only one param to pass to the required commands and functions.

src/nile/common.py
andrew-fleming: I can totally see that. Sounds good!

src/nile/cli.py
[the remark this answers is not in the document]
andrew-fleming: Mmmm agreed! It's much cleaner using switches like that. Thank you :)

andrew-fleming:
> Looks good @andrew-fleming! I left a couple of small suggestions. And I want to add that `watch` sounds better to me than `track` as the option for connecting the terminal waiting for status changes, just for consideration.

@ericnordelo hmm I'm iffy on the suggestion. 

`nile deploy contract --track`

or

`nile deploy contract --watch`

I've been going back and forth. IMO `track` is a more active verb, but `watch` seems more precise. And I agree that the latter sounds better. I'll plug in `watch` and we can reevaluate if needed

src/nile/core/deploy.py
```diff
@@ -22,5 +31,8 @@ def deploy(contract_name, arguments, network, alias, overriding_path=None, abi=N
     )
     logging.info(f"🧾 Transaction hash: {hex(tx_hash)}")
 
+    if status_type is not None:
+        status(tx_hash, network, status_type)
```
ericnordelo, line 35: I think may be a good idea to do this after the full execution of the command (at the end of CLI method, for example) because if the user cancels the execution in the terminal, the accounts and deployments files are not going to be updated, potentially making the user need to re-deploy, or to manually add things (poor UX), even when the transaction was successfully sent.

The suggestion is to block the execution only after the complete execution of the command (this can be better tested in goerli where there is a delay instead of devnet).

ericnordelo (commented):
Hi Andrew, updates look great! I don't have a strong case to defend watch vs track, it is just kind of a feeling, so track is more than good :). I left a suggestion regarding blocking the execution of the command in the middle of it.

src/nile/core/deploy.py
andrew-fleming: Good point! My reservation with moving the status check/command outside of the core lies in the fact that we'd need to repeat the same steps in the NRE (assuming that we want this functionality in the NRE)—open to opinions here.

Otherwise, an easy solution could be to place the status call _after_ deployment registration:

```
# core/deploy.py

def deploy(...):
    ...

    deployments.register(...)

    if status_type is not None:
        status(...)

    return address, register_abi
```

trevis-dev: > Good point! My reservation with moving the status check/command outside of the core lies in the fact that we'd need to repeat the same steps in the NRE (assuming that we want this functionality in the NRE)—open to opinions here.
>

I think it is most valuable to support it in the NRE if users deploy within it.

> Otherwise, an easy solution could be to place the status call _after_ deployment registration:
>

In that case, it shouldn't be too difficult to *un*register on failed transactions either.

andrew-fleming: > I think it is most valuable to support it in the NRE if users deploy within it.

Agreed^

> In that case, it shouldn't be too difficult to unregister on failed transactions either.

Yes, that's a great idea!

ericnordelo (dismissed):
Looking good to me!

src/nile/cli.py
```diff
@@ -88,27 +88,42 @@ def run(path, network):
 @click.argument("arguments", nargs=-1)
 @network_option
 @click.option("--alias")
+@click.option("--track", "-t", "status_type", flag_value="track")
+@click.option("--debug", "-d", "status_type", flag_value="debug")
```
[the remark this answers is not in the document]
martriay: i think i like `watch` better :P @ericnordelo

src/nile/utils/debug.py
```diff
@@ -1,84 +1,61 @@
...
+    return output.decode()
 
-    logging.info(f"🧾 Error message:\n{output.decode()}")
-    return output
+
+def get_addresses_from_string(string):
+    """Return a set of integers with identified addresses in a string."""
+    return set(
+        int(address, 16) for address in re.findall("0x[\\da-f]{1,64}", str(string))
+    )
```
martriay, line 50: this is `common` material

src/nile/utils/debug.py
```diff
@@ -87,6 +64,9 @@ def _abi_to_build_path(filename):
 
 def _locate_error_lines_with_abis(file, addresses, to_contract):
     contracts = []
+    if not os.path.isfile(file):
+        raise IOError(f"Contracts file {file} not found.")
```
martriay, line 68: in what context can this happen?

src/nile/utils/status.py
```diff
@@ -0,0 +1,100 @@
...
+)
+
+
+def status(
+    tx_hash, network, status_type=None, contracts_file=None
+) -> TransactionStatus:
+    """Fetch a transaction status.
+
+    Optionally track until resolved (accepted on L2 or rejected) and/or
+    use available contracts to help locate the error. Debug implies track.
```
martriay, line 24:
```suggestion
    use available artifacts to help locate the error. Debug implies track.
```
I think this is more clear

src/nile/utils/status.py
```diff
@@ -0,0 +1,100 @@
...
+
+    logging.info("⏳ Querying the network for transaction status...")
+
+    receipt = _get_tx_receipt(tx_hash, command, status_type)
+
+    if not receipt.status.is_rejected:
+        return TransactionStatus(tx_hash, receipt.status, None)
+
+    error_message = receipt.receipt["tx_failure_reason"]["error_message"]
+    if status_type == "debug":
```
martriay, line 37: any reason for this not to be the default? when wouldn't you want `debug` enabled?

src/nile/utils/status.py
```diff
@@ -0,0 +1,100 @@
...
+
+        if receipt.status.is_accepted:
+            logging.info(f"✅ {log_output}. No error in transaction.")
+            return receipt
+
+        if status_type is None:
+            logging.info(f"🕒 {log_output}.")
+            return receipt
+
+        logging.info(f"🕒 {log_output}. Trying again in a moment...")
```
martriay, line 69:
```suggestion
        logging.info(f"🕒 {log_output}. Trying again in {RETRY_AFTER_SECONDS} seconds...")
```

martriay (commented):
Looking very good! Left some comments

src/nile/cli.py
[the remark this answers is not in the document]
martriay:
```
declare(...) // defaults to watch_mode="status"
deploy(..., watch_mode="status")
send(..., watch_mode="debug")
```

src/nile/utils/status.py
ericnordelo: In my opinion, I don't want my terminal to be blocked for a couple of minutes. I usually want to send the transaction and check it later, not block the terminal where I'm working (I could open more terminals, but this is not what I expect as a regular flow as a developer).

src/nile/cli.py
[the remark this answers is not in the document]
andrew-fleming: > i think i like watch better :P @ericnordelo

Haha you guys win :) `watch_mode` it is!

README.md
```diff
@@ -278,15 +288,35 @@ Print out the Nile version
...
+Prints the current status of a transaction.
 
 ```sh
-nile debug <transaction_hash> [CONTRACTS_FILE, NETWORK]
+nile status <transaction_hash> [CONTRACTS_FILE, NETWORK, TRACK, DEBUG]
+⏳ Querying the network for transaction status...
+✅ Transaction status: ACCEPTED ON L1. No error in transaction.
+```
+
+#### `--track`, `-t` flag
```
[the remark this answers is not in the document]
ericnordelo: Linking the same comment I made for a similar following comment:
https://github.com/OpenZeppelin/nile/pull/122#discussion_r1002593840

src/nile/utils/status.py
andrew-fleming: I considered setting it as default but reconsidered with the same point @ericnordelo expressed^. I think opt-in seems more appropriate...perhaps, we could include an `auto-debug` feature in the future?

README.md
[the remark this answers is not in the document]
martriay: i can understand you don't want to block the terminal after sending a transaction, but when explicitly requesting the status you want to know whether a tx is confirmed or rejected, i expect the primary use case therefore to be waiting for a resolution when status is pending. if you want to continue, yo can `ctrl + c`

src/nile/utils/debug.py
andrew-fleming: If a user accidentally deletes/renames the deployments.txt file that's being used for the debug query.

README.md
```diff
@@ -278,15 +288,35 @@ Print out the Nile version
...
+🕒 Transaction status: RECEIVED. Trying again in a moment...
+🕒 Transaction status: PENDING. Trying again in a moment...
+✅ Transaction status: ACCEPTED_ON_L2. No error in transaction.
 ```
 
-For example, this transaction returns the very cryptic error message:  
+#### `--debug`, `-d` flag
+
+Use locally available contracts to make error messages from rejected transactions more explicit.  
+NB: Implies `--track`.
```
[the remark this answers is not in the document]
andrew-fleming: Nota Bene (Latin for "note well"). I changed it to "Note"

README.md
```diff
@@ -118,11 +118,12 @@ nile deploy contract --alias my_contract
 
 A few things to notice here:
 
-1. `nile deploy <contract_name>` looks for an artifact with the same name
-2. This created a `localhost.deployments.txt` file storing all data related to my deployment
-3. The `--alias` parameter lets me create a unique identifier for future interactions, if no alias is set then the contract's address can be used as identifier
-4. By default Nile works on local, but you can use the `--network` parameter to interact with `mainnet`, `goerli`, and the default `localhost`.
-5. By default, the ABI corresponding to the contract will be registered with the deployment. To register a different ABI file, use the `--abi` parameter.
+- `nile deploy <contract_name>` looks for an artifact with the same name.
+- This created a `localhost.deployments.txt` file storing all data related to my deployment.
```
ericnordelo, line 122: We should maintain consistency in grammar for the README: 

```suggestion
- This creates or updates the` localhost.deployments.txt` file storing all data related to deployments.
```

There are more places where this happens, like the next sentence, where instead of "lets me create" I would say something like "allows creating".

I think we should maintain third-person singular, but if not, at least we should be consistent.

I would prioritize an issue for refactoring deeply the README for this, or the next milestone @martriay @andrew-fleming .

src/nile/cli.py
```diff
@@ -103,15 +120,18 @@ def declare(
...
 
 @cli.command()
 @click.argument("signer", nargs=1)
 @network_option
-def setup(signer, network):
+@click.option("--track", "-t", "watch_mode", flag_value="track")
+@click.option("--debug", "-d", "watch_mode", flag_value="debug")
+def setup(signer, network, watch_mode):
     """Set up an Account contract."""
-    Account(signer, network)
```
ericnordelo, line 114: This is probably for a different issue, but I think we should refactor the account to avoid deploying when the instance is being initialized and make the deploy step explicit by calling deploy if necessary. I think this would make the code clearer.

src/nile/cli.py
```diff
@@ -120,10 +140,21 @@ def setup(signer, network):
 @click.argument("method", nargs=1)
 @click.argument("params", nargs=-1)
 @click.option("--max_fee", nargs=1)
+@network_option
+@click.option("--track", "-t", "watch_mode", flag_value="track")
```
ericnordelo, line 144: Should we compress both pairs (watch_mode and query) into a single option (like network_option) to reduce unnecessary verbosity in the code?

src/nile/cli.py
```diff
@@ -233,8 +273,39 @@ def version():
...
+    Get the status of a transaction.
+
+    $ nile status transaction_hash
+      Get the current status of a transaction.
+
+    $ nile status --track transaction_hash
+      Get (wait for) the final status of a transaction (REJECTED / ACCEPTED ON L2)
+
+    $ nile status --debug transaction_hash
+      Same as `status --track` then locate errors if rejected using local contracts
```
ericnordelo, line 301:
```suggestion
      Same as `status --track` then locate errors if rejected using local artifacts
```

To keep consistency with Marto's suggestion.

src/nile/utils/status.py
```diff
@@ -0,0 +1,100 @@
+"""Functions used to find/track/debug a transaction status."""
```
ericnordelo, line 1: For me, this file looks weird under the utils folder. I think we should check the issue for organizing the code structure soon because is going to be harder while the codebase continues growing.

This would loog better in "core/commands/helpers" IMO.

src/nile/utils/status.py
```diff
@@ -0,0 +1,100 @@
...
+            logging.info(f"🕒 {log_output}.")
+            return receipt
+
+        logging.info(
+            f"🕒 {log_output}. Trying again in {RETRY_AFTER_SECONDS} seconds..."
+        )
+        time.sleep(RETRY_AFTER_SECONDS)
+
+
+class Status(Enum):
```
ericnordelo, line 73: For clarity, I would name this TxStatus (or something different than just Status). When this is imported in deploy.py for example, this line: `from nile.utils.status import Status, status` looks a little confusing.

```suggestion
class TxStatus(Enum):
```

ericnordelo (commented):
Looking good! I left some comments.

README.md
andrew-fleming: Mmm yeah, I'm not a fan of the first-person. I'd say third-person singular or even second person ("lets you create"). I agree on prioritizing a README refactor.

src/nile/core/declare.py
```diff
@@ -42,6 +44,13 @@ def declare(
     logging.info(f"🧾 Transaction hash: {hex(tx_hash)}")
 
     deployments.register_class_hash(class_hash, network, alias)
+
+    if watch_mode is not None:
+        is_accepted = status(tx_hash, network, watch_mode)
+        if is_accepted.status == Status.REJECTED:
```
trevismd, line 50: You guys are doing great, thank you for giving more and more depth to the PR!
Here, I'd like to say that the `Status` class itself was not meant to be exported (for the `[...] import Status, status` situation).

Instead, these line could be written as
 
```suggestion
        tx_status = status(tx_hash, network, watch_mode)
        if tx_status.status.is_rejected:
```
or even
```suggestion
        if status(tx_hash, network, watch_mode).status.is_rejected:
```
which allows to keep the implementation of is_rejected separate.

README.md
ericnordelo: I saw now I wrote third-person singular where I meant third-person present, sorry.

src/nile/core/declare.py
```diff
@@ -42,6 +44,13 @@ def declare(
...
+            deployments.unregister(class_hash, network, alias, is_deployment=False)
```
trevis-dev, line 51: You guys are doing great, thank you for giving more and more depth to the PR!
Here, I'd like to say that the `Status` class itself was not meant to be exported (for the `[...] import Status, status` situation).

Instead, these line could be written as
 
```suggestion
        tx_status = status(tx_hash, network, watch_mode)
        if tx_status.status.is_rejected:
```
or even
```suggestion
        if status(tx_hash, network, watch_mode).status.is_rejected:
```
which allows to keep the implementation of is_rejected separate.

src/nile/core/declare.py
trevis-dev, line 50: You guys are doing great, thank you for giving more and more depth to the PR!
Here, I'd like to say that the `Status` class itself was not meant to be exported (for the `[...] import Status, status` situation).

Instead, these line could be written as
 
```suggestion
        tx_status = status(tx_hash, network, watch_mode)
        if tx_status.status.is_rejected:
```
or even
```suggestion
        if status(tx_hash, network, watch_mode).status.is_rejected:
```
which allows to keep the implementation of is_rejected separate. 

(last edit ;-) )

README.md
andrew-fleming: Haha I rewrote the same thing. I understood what you meant :)

src/nile/core/declare.py
andrew-fleming: Thank you, @trevis-dev! Yes, good call. I think the second suggestion looks cleaner^

src/nile/cli.py
andrew-fleming: I think it makes more sense to do so with `watch_mode` as opposed to `query` because we only insert `query` into `account.send`. I'm happy to add it if think otherwise.

README.md
```diff
@@ -118,29 +118,12 @@ nile deploy contract --alias my_contract [TRACK, DEBUG]
 
 A few things to notice here:
```
martriay, line 119:
```suggestion
A few things to note here:
```
this is better right?

README.md
```diff
@@ -118,29 +118,12 @@ nile deploy contract --alias my_contract [TRACK, DEBUG]
...
-4. By default Nile works on local, but you can use the `--network` parameter to interact with `mainnet`, `goerli`, and the default `localhost`.
-5. `--track` and `--debug` flags : Chain `declare` call with `nile status` + the chosen flag. See `status` below for a complete description.
+- `nile deploy <contract_name>` looks for an artifact with the same name.
+- This creates or updates the` localhost.deployments.txt` file storing all data related to deployments.
+- The `--alias` parameter creates a unique identifier for future interactions, if no alias is set then the contract's address can be used as identifier.
+- By default Nile works on local, but you can use the `--network` parameter to interact with `mainnet`, `goerli`, and the default `localhost`.
+- By default, the ABI corresponding to the contract will be registered with the deployment. To register a different ABI file, use the `--abi` parameter.
+- `--track` and `--debug` flags can be used to chain the `setup` call with `nile status` + the chosen flag. See `status` below for a complete description.
```
martriay, line 126:
```suggestion
- `--track` and `--debug` flags can be used to watch the status of the deployment transaction. See `status` below for a complete description.
```

README.md
```diff
@@ -161,16 +144,17 @@ nile setup <private_key_alias> [TRACK, DEBUG]
 
 A few things to note here:
 
-1. `nile setup <private_key_alias>` looks for an environment variable with the name of the private key alias
-2. This creates a `localhost.accounts.json` file storing all data related to accounts management
-3. `--track` and `--debug` flags : Chain `declare` call with `nile status` + the chosen flag. See `status` below for a complete description.
+- `nile setup <private_key_alias>` looks for an environment variable with the name of the private key alias.
+- This creates or updates `localhost.accounts.json` file storing all data related to accounts management.
+- The creates or updates `localhost.deployments.txt` file storing all data related to deployments.
+- `--track` and `--debug` flags can be used to chain the `setup` call with `nile status` + the chosen flag. See `status` below for a complete description.
```
martriay, line 150:
```suggestion
- `--track` and `--debug` flags can be used to watch the status of the account deployment transaction. See `status` below for a complete description.
```

README.md
```diff
@@ -185,27 +169,31 @@ Transaction hash: 0x1c
...
 Some things to note:
 
+- This sends the transaction to the network by default, but you can use the `--estimate_fee` flag to estimate the fee without sending the transaction, or the `--simulate` flag to get a traceback of the simulated execution.
 - `max_fee` defaults to `0`. Add `--max_fee <max_fee>` to set the maximum fee for the transaction.
 - `network` defaults to the `localhost`. Add `--network <network>` to change the network for the transaction.
-
-#### `--track` and `--debug` flags
-
-Chain `send` calls with `nile status` + the chosen flag. See `status` below for a complete description.
+- `--track` and `--debug` flags can be used to chain the `send` call with `nile status` + the chosen flag. See `status` below for a complete description.
```
martriay, line 175:
```suggestion
- `--track` and `--debug` flags can be used to watch the status of the transaction. See `status` below for a complete description.
```

README.md
```diff
@@ -185,27 +169,31 @@ Transaction hash: 0x1c
...
-2. This creates or updates a `localhost.declarations.txt` file storing all data related to your declarations
-3. The `--alias` parameter lets you create a unique identifier for future interactions, if no alias is set then the contract's address can be used as identifier
-4. By default Nile works on local, but you can use the `--network` parameter to interact with `mainnet`, `goerli`, and the default `localhost`.
+- `nile declare <private_key_alias> <contract_name>` looks for an artifact with name `<contract_name>`.
+- This creates or updates a `localhost.declarations.txt` file storing all data related to your declarations.
+- The `--alias` parameter lets you create a unique identifier for future interactions, if no alias is set then the contract's address can be used as identifier.
+- By default Nile works on local, but you can use the `--network` parameter to interact with `mainnet`, `goerli`, and the default `localhost`.
+- `--track` and `--debug` flags can be used to chain the `declare` call with `nile status` + the chosen flag. See `status` below for a complete description.
```
martriay, line 196:
```suggestion
- `--track` and `--debug` flags can be used to watch the status of the declaration transaction. See `status` below for a complete description.
```

README.md
```diff
@@ -278,7 +262,7 @@ nile version
 Prints the current status of a transaction.
 
 ```sh
-nile status <transaction_hash> [CONTRACTS_FILE, NETWORK, TRACK, DEBUG]
+nile status <transaction_hash> [CONTRACTS_FILE, NETWORK, --track, --debug]
```
martriay, line 265: i think this means you can use both together

src/nile/cli.py
```diff
@@ -125,53 +144,39 @@ def setup(signer, network, status_type):
...
+@click.option("--estimate_fee", "query", flag_value="estimate_fee")
+def send(
+    signer,
+    address_or_alias,
+    method,
+    params,
+    network,
+    watch_mode,
+    max_fee=None,
+    query=None,
```
martriay, line 159: are these default params needed in click? if they are, i'd add them to `watch_mode`, otherwise we can remove them no?

src/nile/core/account.py
```diff
@@ -85,26 +85,16 @@ def deploy(self, status_type=None):
 
         return address, index
 
-    def send(
+    def declare(
```
martriay, line 88: wat there were two `send`?

src/nile/deployments.py
```diff
@@ -28,10 +28,18 @@ def register(address, abi, network, alias):
         fp.write("\n")
 
 
-def unregister(address_or_class_hash, network, alias, abi=None):
+def unregister(address_or_class_hash, network, alias, abi=None, is_deployment=True):
```
martriay, line 31:
```suggestion
def unregister(address_or_class_hash, network, alias, abi=None, is_declaration=False):
```
since the flag will be passed for declarations, i think we should name it like this for clarity

src/nile/deployments.py
```diff
@@ -28,10 +28,18 @@ def register(address, abi, network, alias):
...
-    to_delete = hex_address(address_or_class_hash)
+    file = (
+        f"{network}.{DEPLOYMENTS_FILENAME}"
+        if is_deployment
+        else f"{network}.{DECLARATIONS_FILENAME}"
+    )
+    to_delete = (
+        hex_address(address_or_class_hash)
+        if is_deployment
+        else hex(address_or_class_hash)
```
martriay, line 41: the difference being padding?

martriay (commented):
Looking very good!

src/nile/core/account.py
andrew-fleming: Git thinks so! Part of the merging fun :)

src/nile/deployments.py
andrew-fleming: Yes, indeed

src/nile/cli.py
andrew-fleming: Good catch, they are not needed

README.md
```diff
@@ -109,20 +109,21 @@ Creating artifacts/abis/ to store compilation artifacts
 > NOTICE: this method doesn't use an account, which will be deprecated very soon as StarkNet makes deployments from accounts mandatory.
 
 ```sh
-nile deploy contract --alias my_contract
+nile deploy contract --alias my_contract [TRACK/DEBUG]
```
martriay, line 112:
```suggestion
nile deploy contract --alias my_contract [--track/--debug]
```
i'm not sure i'm familiar with that syntax, shouldn't it be like this?

src/nile/core/call_or_invoke.py
```diff
@@ -53,3 +58,16 @@ def call_or_invoke(
         max_fee=max_fee,
         query_flag=query_flag,
     )
+
+    if type != "call" and output:
+        logging.info(output)
+        if not query_flag:
```
martriay, line 64: shouldn't it also check whether `watch_mode=False`?

src/nile/core/declare.py
```diff
@@ -42,6 +44,12 @@ def declare(
     logging.info(f"🧾 Transaction hash: {hex(tx_hash)}")
 
     deployments.register_class_hash(class_hash, network, alias)
+
+    if watch_mode is not None:
+        if status(tx_hash, network, watch_mode).status.is_rejected:
+            deployments.unregister(class_hash, network, alias, is_declaration=True)
```
martriay, line 50: this makes me think the default should be `watch_mode=debug`. otherwise it's easy for users to enter an invalid state (failed deployment being registered)

src/nile/deployments.py
```diff
@@ -28,6 +28,33 @@ def register(address, abi, network, alias):
         fp.write("\n")
 
 
+def unregister(address_or_class_hash, network, alias, abi=None, is_declaration=False):
+    """Unregister deployment or class hash from file."""
+    file = (
+        f"{network}.{DEPLOYMENTS_FILENAME}"
+        if not is_declaration
+        else f"{network}.{DECLARATIONS_FILENAME}"
```
martriay, line 36:
```suggestion
        f"{network}.{DECLARATIONS_FILENAME}"
        if is_declaration
        else f"{network}.{DEPLOYMENTS_FILENAME}"
```

src/nile/deployments.py
```diff
@@ -28,6 +28,33 @@ def register(address, abi, network, alias):
...
+    )
+    to_delete = (
+        hex_address(address_or_class_hash)
+        if not is_declaration
+        else hex(address_or_class_hash)
```
martriay, line 41:
```suggestion
        hex(address_or_class_hash)
        if is_declaration
        else hex_address(address_or_class_hash)
```

martriay (commented):
we're so close 🥺

martriay:
fix conflicts again please 😢

martriay (dismissed):

README.md
```diff
@@ -262,15 +266,35 @@ Print out the Nile version
...
 
-### `debug`
+### `status`
 
-Use locally available contracts to make error messages from rejected transactions more explicit.
+Prints the current status of a transaction.
 
 ```sh
-nile debug <transaction_hash> [CONTRACTS_FILE, NETWORK]
+nile status <transaction_hash> [CONTRACTS_FILE, NETWORK, TRACK/DEBUG]
```
martriay, line 274:
```suggestion
nile status <transaction_hash> [CONTRACTS_FILE, NETWORK, TRACK/DEBUG]
```
these examples are inconsistent with the recent changes (`--track/--debug`). i'm still a bit confused by this syntax. is it common to refer to flags like this?

martriay: because i see it used for other flags/params as well

andrew-fleming: I'm honestly not sure if it's common. I'll update the examples to something less confusing

README.md
```diff
@@ -86,6 +86,12 @@ WARNING: Use these accounts and their keys ONLY for local testing. DO NOT use th
 
 Compile Cairo contracts. Compilation artifacts are written into the `artifacts/` directory.
 
+```txt
+nile compile [PATH_TO_CONTRACT | --directory DIRECTORY] [--disable-hint-validation]
```
martriay, line 90:
```suggestion
nile compile [PATH_TO_CONTRACT] [--directory DIRECTORY] [--disable-hint-validation]
```
`PATH_TO_CONTRACT` and `--directory` are not really alternatives aren't they?

README.md
```diff
@@ -150,16 +161,17 @@ nile setup <private_key_alias>
...
+- The creates or updates `localhost.deployments.txt` file storing all data related to deployments.
+- `--track` and `--debug` flags can be used to watch the status of the account deployment transaction. See `status` below for a complete description.
 
 ### `send`
 
 Execute a transaction through the `Account` associated with the private key provided. The syntax is:
 
 ```sh
-nile send <private_key_alias> <contract_identifier> <method> [PARAM_1, PARAM2...]
+nile send [--track | --debug] <private_key_alias> <contract_identifier> <method> [PARAM_1, PARAM2...]
```
martriay, line 174:
```suggestion
nile send <private_key_alias> <contract_identifier> <method> [PARAM_1, PARAM2...] [--track | --debug]
```

src/nile/core/declare.py
martriay: yes i think it should

martriay (approved):
great work :)

src/nile/core/declare.py
andrew-fleming: Haha I don't disagree on that point. Okay, let's give `watch_mode=debug` a trial run. If it doesn't go over well, we can always change it

README.md
andrew-fleming: To my eyes, when you add `PATH_TO_CONTRACT`, `--directory` serves no purpose and vice versa, no?

`nile compile path/to/contract.cairo`

and

`nile compile --directory path/to/directory`

src/nile/core/declare.py
martriay: the alternative is to store the tx_hash. if there's a tx_hash, when someone tries to read the deployment it fetches the status, removing the entry on failure and completing the deployment info on success

andrew-fleming: I don't love the idea of deployment registration being a two-step process for users. I'm not opposed to storing the tx hash (in conjunction with the deployment data) and removing the entry if it's rejected.

Or, what about storing just the hash _with_ `watch_mode=debug`? That will prevent users from sending txs to contracts that haven't been accepted yet (I ran into this error a few times) and, as you mentioned, complete the deployment info on success

README.md
martriay: I believe it also sets the cairo_path for the compiler. for example, this is needed to compile OZ contracts:
```sh
nile compile src/openzeppelin/account/presets/EthAccount.cairo --directory src
```

src/nile/core/declare.py
martriay: what don't you like about the two step process? it can be invisible to the user, since the 2nd step will happen on read.

andrew-fleming: On second thought, I take it back. I can see the two-step process being a good option

src/nile/cli.py
```diff
@@ -88,26 +95,37 @@ def run(path, network):
 @click.argument("arguments", nargs=-1)
 @network_option
 @click.option("--alias")
+@watch_option
 @click.option("--abi")
 @mainnet_token_option
-def deploy(artifact, arguments, network, alias, abi=None, token=None):
+def deploy(artifact, arguments, network, alias, watch_mode, abi=None, token=None):
```
ericnordelo, line 101: I think we can remove the initialization in Click command methods, because the default behavior for options is passing None.

src/nile/cli.py
```diff
@@ -88,26 +95,37 @@ def run(path, network):
 @click.argument("arguments", nargs=-1)
```
ericnordelo, line 95: A small suggestion: wouldn't be good to organize the decorators of click commands (by size or type?). Params of the method don't need to match the order of the decorator because Click's wrapper is the one passing the argument by name.

src/nile/utils/debug.py
```diff
@@ -1,84 +1,53 @@
...
+def debug_message(error_message, command, network, contracts_file=None):
+    """
+    Use available contracts to help locate the error in a rejected transaction.
+
+    :param error_message: error message of the transaction receipt.
+    :param command: StarkNet CLI command used to get the transaction receipt.
+    :param network: Network queried
+    :param contracts_file: File to use instead of the one generated automatically
+      from network name.
+    """
```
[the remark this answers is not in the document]
ericnordelo: Just to keep consistency. If we agree to use the :param format is ok with me.

ericnordelo (commented):
Looking good Andrew! Left a couple of small suggestions.

src/nile/cli.py
andrew-fleming: Haha yeah, I agree it'd be a good idea to better organize the decorators. I was matching the order because it seemed practical, but I think it looks a little weird—especially with the different `@_option`s.

src/nile/deployments.py
```diff
@@ -28,6 +28,33 @@ def register(address, abi, network, alias):
...
 
+def unregister(address_or_class_hash, network, alias, abi=None, is_declaration=False):
+    """Unregister deployment or class hash from file."""
+    file = (
+        f"{network}.{DECLARATIONS_FILENAME}"
+        if is_declaration
+        else f"{network}.{DEPLOYMENTS_FILENAME}"
+    )
+    to_delete = (
+        hex(address_or_class_hash)
```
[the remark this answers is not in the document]
andrew-fleming: Yes, good call!

src/nile/utils/debug.py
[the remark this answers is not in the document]
andrew-fleming: Right! No strong opinions on formatting here. I'll add yours for now for consistency

martriay (approved):

ericnordelo (approved):
Looking good to go!

PR #150, opened 2022-07-24
Fix passing optional args in `_compile_contract`

### LINKED ISSUE #149: Error combining `--account_contract` and `--disable_hint_validation` arguments
Attempting to combine both attributes does not work as they are not separated by a space when passed to `starknet-compile`:

```
nile compile --account_contract --disable-hint-validation MyAccount.cairo
```

throws error:

```
starknet-compile: error: unrecognized arguments: --account_contract--disable_hint_validation
```

andrew-fleming:
When executing `nile compile` with the optional `--account_contract` and `--disable_hint_validation` flags, `_compile_contract` concatenated the flags i.e. `--account_contract--disable_hint_validation`.

Resolves #149.

### DIFF
--- a/src/nile/core/compile.py
+++ b/src/nile/core/compile.py
@@ -67,10 +67,10 @@ def _compile_contract(
     """
 
     if account_contract or filename.endswith("Account"):
-        cmd = cmd + "--account_contract"
+        cmd = cmd + "\n--account_contract"
 
     if disable_hint_validation:
-        cmd = cmd + "--disable_hint_validation"
+        cmd = cmd + "\n--disable_hint_validation"
 
     process = subprocess.Popen(cmd.split(), stdout=subprocess.PIPE)
     process.communicate()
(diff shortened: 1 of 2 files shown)


src/nile/core/compile.py
```diff
@@ -67,10 +67,10 @@ def _compile_contract(
     """
 
     if account_contract or filename.endswith("Account"):
-        cmd = cmd + "--account_contract"
+        cmd = cmd + "\n--account_contract"
```
martriay, line 70:
```suggestion
        cmd.append("--account_contract")
```
isn't this better? i'm basing it on the tests

martriay (approved):
Looks good! left a small suggestion

src/nile/core/compile.py
andrew-fleming: It would be, but `cmd` is written as a string (which makes sense since they'd all be f-strings anyway):

```
    cmd = f"""
    starknet-compile {path} \
        --cairo_path={contracts_directory}
        --output {BUILD_DIRECTORY}/{filename}.json \
        --abi {ABIS_DIRECTORY}/{filename}.json
    """
```

martriay: right!

martriay (approved):

PR #144, opened 2022-07-11
Add lite mode to node run

JulissaDantes:
Fixes #116 .

Includes the lite mode parameter on the node run,  an addition to the test_node test,and for testing purposes the account deploy returns the transaction hash.

*Note I'll be a draft until the issues that came after the merge are solved

PR #153, opened 2022-07-25
Add lite-mode

### LINKED ISSUE #116: Add lite mode param to devnet
https://github.com/Shard-Labs/starknet-devnet#lite-mode

andrew-fleming:
This PR supercedes #144 and resolves #116.


README.md
```diff
@@ -46,15 +46,34 @@ This command creates the project directory structure and installs `cairo-lang`,
...
+nile node [--host HOST] [--port PORT] [--lite_mode]
+
+optional arguments:
+--host HOST         Specify the address to listen at; defaults to
+                    127.0.0.1 (use the address the program outputs on
+                    start)
+--port PORT         Specify the port to listen at; defaults to 5050
+--lite-mode         Applies all lite-mode optimizations by disabling
+                    features such as block hash calculation and deploy
+                    hash
```
martriay, line 59:
```suggestion
                    features such as block hash and deploy hash
                    calculation
```

tests/commands/test_node.py
```diff
@@ -0,0 +1,48 @@
...
+    monkeypatch.chdir(tmp_path)
+    return tmp_path
+
+
+@pytest.mark.parametrize(
+    "args, host, port, mode",
+    [
+        ([], HOSTS[0], PORTS[0], None),
+        ([HOSTS[1], PORTS[1]], HOSTS[1], PORTS[1], None),
+        ([HOSTS[1], PORTS[1], LITE], HOSTS[1], PORTS[1], LITE),
```
martriay, line 25: i’m not sure why these two lines use `HOSTS[1]`, `PORTS[1]`, and `LITE` as args. i guess they’re just placeholder values but if that's the case then it causes a bit of confusion. or do they serve a different purpose?

tests/commands/test_node.py
```diff
@@ -0,0 +1,48 @@
...
+    ],
+)
+@patch("nile.core.node.subprocess.check_call")
+def test_node_call(mock_subprocess, args, host, port, mode):
+    node(*args)
+
+    command = ["starknet-devnet", "--host", host, "--port", port]
+    if mode:
+        command.append(mode)
+    mock_subprocess.assert_called_once_with(command)
```
martriay, line 35: aren't we testing `args`?

martriay (commented):
Looking good! Left some comments

tests/commands/test_node.py
andrew-fleming: It's to make sure that when these args are explicitly passed, it doesn't cause any issues with the command (like it did with `--account_contract` and `disable_hint_validation ` in #150). I'll refactor to make it less confusing

tests/commands/test_node.py
andrew-fleming: Yes! I think the weirdness of this test comes from checking the defaults (which differ from args except when explicitly set)

martriay (approved):

Issue #117, opened 2022-05-09
Improve error bubbling


martriay:
As seen in many issues as https://github.com/OpenZeppelin/nile/issues/108, when there's an upstream error in Nile (e.g. when the underlying command such as `starknet` or `starknet-devnet` fails), the output is very obscure about the reasons of the error.

Fix nile so it propagates the error message and help users troubleshoot nile.

fracek:
I think the way to do this is by capturing the subprocess output (should we differentiate stdout and stderr with e.g. color?), capture any exception to suppress the default python stacktrace and simply inform the user something went wrong.

PR #158, opened 2022-08-02
Improve error handling of account init with invalid key

andrew-fleming:
Contributes toward #117.

### DIFF
--- a/src/nile/core/account.py
+++ b/src/nile/core/account.py
@@ -1,4 +1,5 @@
 """Command to call or invoke StarkNet smart contracts."""
+import logging
 import os
 
 from dotenv import load_dotenv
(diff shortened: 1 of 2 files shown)


src/nile/core/account.py
```diff
@@ -20,8 +21,16 @@ class Account:
...
-        self.signer = Signer(int(os.environ[signer]))
-        self.network = network
+        try:
+            self.signer = Signer(int(os.environ[signer]))
+            self.network = network
+        except KeyError:
+            logging.error(
+                f"\n❌ Cannot find {signer} in env."
+                "\nCheck spelling and that it exists."
+                "\nTry moving the .env to the directory outside of your project."
```
martriay, line 31: where does this specific recommendation come from?

martriay (approved):
Looks good! Just left a question.

src/nile/core/account.py
andrew-fleming: When I first started using Nile, this happened to me. And I recently walked a friend through using Nile, and he also had his .env inside the project. I think it might be a common-enough problem that we can anticipate

martriay: I always have it inside the project, never had any issues. What kind of issue is that?

andrew-fleming: It's the `KeyError` issue (i.e. not found)...second thought, `project` might not be the correct word there^. When I say `project`, I mean the local Nile repo. Maybe `Try moving the .env to the root of your project` is a better recommendation

martriay: please do!

martriay (approved):

PR #160, opened 2022-08-03
Add no-wallet to account command

andrew-fleming:
Meant to be included in #154.


martriay (approved):
Heh just noticed the same. Thanks for fixing it.

PR #162, opened 2022-08-11
Sanitize call/invoke params

martriay:
This PR aims to sanitize parameters sent to `nre` functions like `deploy`, `call`, and `invoke` as well as `account.send`. For this, it recursively converts parameters to strings in `common.run_command` and `call_or_invoke`.

It also `.split()`s the return value of `call` and `invoke` operations to better parse return values (e.g. `uint`)

### DIFF
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -8,7 +8,7 @@ from nile.common import BUILD_DIRECTORY, run_command
 CONTRACT = "contract"
 OPERATION = "invoke"
 NETWORK = "goerli"
-ARGS = [1, 2, 3]
+ARGS = ['1', '2', '3']
 
 
 @pytest.mark.parametrize("operation", ["invoke", "call"])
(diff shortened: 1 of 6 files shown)


tests/test_common.py
```diff
@@ -8,7 +8,7 @@
 CONTRACT = "contract"
 OPERATION = "invoke"
 NETWORK = "goerli"
-ARGS = [1, 2, 3]
+ARGS = ["1", "2", "3"]
```
andrew-fleming, line 11: I'm wondering if we should keep the `ARGS` as ints and assert they come out as strings in the subprocess. Maybe it's not necessary fi we isolate and test `stringify`

andrew-fleming (commented):
Much cleaner, well done! I think we should add tests for `stringify` with nested lists

tests/test_common.py
martriay: i prefer the shortest path at this point 😅

andrew-fleming (approved):

Issue #165, opened 2022-08-13
utils imports block `nile init`


andrew-fleming:
The recently added utils (`nile.utils.__init__`) includes imports from cairo-lang. Since `nile init` installs cairo-lang, and utils declares imports from the yet-to-be-installed cairo-lang, this command is blocked.

Issue #134, opened 2022-06-28
Dependency management is not accurate


ClementWalter:
I'm quite confused with how `nile` expects the user to handle dependencies management.
Nile is not a dependency management tool itself though `nile init` does run some installation that are, however, not given in package extra:

```
[package.dependencies]
click = ">=8.0,<9.0"
importlib-metadata = ">=4.0"
python-dotenv = ">=0.19.2"

[package.extras]
testing = ["cairo-lang", "setuptools", "tox", "twine", "pytest", "pytest-cov"]
```

while the `README` says:

> This command creates the project directory structure and installs cairo-lang, starknet-devnet, pytest, and pytest-asyncio for you.

I thinks that this pattern creates confusion between tools and that in usual python packaging, if a given package requires some extra, the proper way to have them is *not* to hard install them in a `init` command *but* to give a proper list of extras.

An alternative solution would be to edit the pyproject.toml file with the added deps as would, for example, `hardhat` do with `package.json`.

martriay:
Sounds right. Could you open a PR for it?

ClementWalter:
would you go for editing the `pyproject.toml` or adding the deps in extra? My thought are:
- because pyproject.toml is not the only way to manage deps in python, this is not as general as package.json (like this project uses steup.cfg)
- adding an extra would make the user always install `nile` with this flag
Eventually I'm questioning this status of being "extra" deps and I guess that they should just be added to the package deps themselves. Also they should have versions requirements as well.

I personally am a huge fan of poetry and I can't see which tool is currently used by nile so I'd recommended migrating to using poetry as well

andrew-fleming:
@ClementWalter because Nile requires the cairo-lang installation, I'm thinking we just move this dep as an install requirement in the setup.cfg. I believe this automatically takes care of most of the deps within `nile init`. Following your advice in this issue, I think we should move the starknet-devnet to extras (it's the only dep not accounted for with this approach). Thoughts?

And regarding poetry, we're looking into it to see if it makes sense. Thank you for the suggestion!

PR #166, opened 2022-08-13
Temp fix for `nile init`

andrew-fleming:
Temporary fix for #165. This should be better addressed in #134.


martriay (approved):
We should make sure we remove this after solving #134

PR #168, opened 2022-08-19
Fix CLI param handling

### LINKED ISSUE #167: Param handling causing CLI `call` and direct `invoke` errors
`stringify` receives a tuple from the CLI instead of the expected list.

andrew-fleming:
This PR proposes to fix the CLI's `call` and `invoke`. This includes:
- applying `stringify` to tuples
  - cli.py passes args as tuples; therefore, `stringify` did not handle these args correctly
- moving the string `split` to the NRE's specific `call` method
  - otherwise, the CLI output returned a list of strings i.e. `["0"]`

Resolves #167.

### DIFF
--- a/src/nile/common.py
+++ b/src/nile/common.py
@@ -80,7 +80,7 @@ def parse_information(x):
 
 
 def stringify(x):
-    """Recursively convert list elements to strings."""
+    """Recursively convert list or tuple elements to strings."""
     if isinstance(x, list) or isinstance(x, tuple):
         return [stringify(y) for y in x]
     else:
--- a/src/nile/common.py
+++ b/src/nile/common.py
@@ -81,7 +81,7 @@ def parse_information(x):
 
 def stringify(x):
     """Recursively convert list elements to strings."""
-    if isinstance(x, list):
+    if isinstance(x, list) or isinstance(x, tuple):
         return [stringify(y) for y in x]
     else:
         return str(x)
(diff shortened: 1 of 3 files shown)


martriay:
do you have an example to reproduce the issue and the fix?

andrew-fleming:
> do you have an example to reproduce the issue and the fix?

Yes! Starting with a fresh env, install the current release of nile and `nile init`. Then:
```
nile node
nile compile src/nile/base_project/contracts/contract.cairo
nile deploy contract --alias my_contract
nile setup <PRIVATE_KEY>
```

- Running `nile call my_contract get_balance` and `nile invoke my_contract increase_balance 123` both return the following error:
`Error: ValueError: Invalid input value: '('. Expected a decimal or hexadecimal integer.`

- Running `nile send PRIVATE_KEY my_contract increase_balance 123` will execute but this is the terminal output:
```
Calling increase_balance on my_contract with params: ['123']
['Invoke', 'transaction', 'was', 'sent.', 'Contract', 'address:', '0x063fe62512433ee8e7ca8d69922883927a582f5cc67aa115c9846bc0fb09f2b1', 'Transaction', 'hash:', '0x7c3346703269a8b5694396ab06b1e2dce62dad36c6bb0226ceeec2bcf456589']
```

The NRE works perfectly using the [cairo-workshop scripts and contracts](https://github.com/martriay/cairo-workshop) as an example.

---

With the proposed changes in this PR, retry the steps outlined above. `nile call` and `nile invoke` should execute and `nile send` doesn't return the command array. Further, run the scripts mentioned above to prove this PR doesn't break the scripting feature.

martriay (approved):

PR #169, opened 2022-08-22
Replace subprocess with direct python calls

### LINKED ISSUE #161: Replace subprocess with direct python calls
Currently `starknet` CLI calls are done through python's `subprocess`, which results in quirky implementations and error handling. These can be replaced with direct python calls through the module system.

andrew-fleming:
This PR proposes to replace subprocess with direct python calls for StarkNet interaction.

Resolves #161.

### DIFF
--- a/src/nile/core/declare.py
+++ b/src/nile/core/declare.py
@@ -2,15 +2,13 @@
 import logging
 
 from nile import deployments
-from nile.common import DECLARATIONS_FILENAME, ABIS_DIRECTORY, BUILD_DIRECTORY, GATEWAYS
+from nile.common import DECLARATIONS_FILENAME, ABIS_DIRECTORY, BUILD_DIRECTORY, get_gateway_response
 from starkware.starknet.services.api.gateway.transaction import Declare, DECLARE_SENDER_ADDRESS
 from starkware.starknet.services.api.contract_class import ContractClass
 from starkware.starknet.definitions import constants
-from starkware.starknet.services.api.gateway.gateway_client import GatewayClient
-from starkware.starkware_utils.error_handling import StarkErrorCode
 
 
-async def declare(contract_name, network, alias=None, overriding_path=None, token=None):
+async def declare(contract_name, network, alias=None, signature=None, overriding_path=None, token=None):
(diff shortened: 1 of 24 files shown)


src/nile/core/declare.py
```diff
@@ -1,24 +1,58 @@
...
+        overriding_path if overriding_path else (BUILD_DIRECTORY, ABIS_DIRECTORY)
+    )
+    artifact = f"{base_path[0]}/{contract_name}.json"
+    open_artifact = open(artifact, "r")
+    contract_class = ContractClass.loads(data=open_artifact.read())
+
+    if signature is None:
+        signature = []
+
+    tx = Declare(
```
ericnordelo, line 40: Why are we sending the transaction at this low level instead of using the [declare](https://github.com/starkware-libs/cairo-lang/blob/0ba3ff59c1c86f2a30adc8fd144eaacb22c48ce9/src/starkware/starknet/cli/starknet_cli.py#L710) function directly?

I think we are trying to address two problems at the same time, the first one being **replacing subprocess calls with direct python calls** (as the name of the PR states), and the second one is **obtaining output in a better format than string** (from print statements).

To solve the first problem, we can import and use the commands from the `starknet.cli` module directly, (like [declare](https://github.com/starkware-libs/cairo-lang/blob/0ba3ff59c1c86f2a30adc8fd144eaacb22c48ce9/src/starkware/starknet/cli/starknet_cli.py#L710)), we don't need to simulate the implementation of that command (as I think we are trying to do with the current code). The current code solves the second problem though, because sadly starknet doesn't provide a Runtime Environment, so to get the output in a better format, we need to "hack" the cli.

Summarizing I think we could replace subprocess more easily by using the CLI commands where we currently call subprocess (I may be missing the full picture). But I think the second problem of course is still super important (could be separated into a different issue maybe?).

andrew-fleming: @ericnordelo I totally agree with the above points^ haha that's why in the new PR (https://github.com/OpenZeppelin/nile/pull/236), we are using the `starknet_cli` directly. I'll close this PR to avoid further confusion. Sorry about that!

PR #175, opened 2022-08-27
Fix binding to all interfaces issue

### LINKED ISSUE #171: Bug: nile node --host 0.0.0.0 creates incorrect node.json file
I'm using docker to run nile, therefore I'm binding the starknet-devnet node to 0.0.0.0 (all interfaces), in order to be able to access the server from my host (with Braavos in browser for example). The command:

`nile node --host 0.0.0.0`

The following snippet shows that in node.py, the "network" is being assigned as **0.0.0.0** instead of **localhost**

https://github.com/OpenZeppelin/nile/blob/bb37ca4cd8effd164bef05d200b38dd835235423/src/nile/core/node.py#L13-L22

and after that the node.json file is created with the following content:

```
{"0.0.0.0": "http://0.0.0.0:5050/"}
```

This results in a posterior error when deploying to the default network "localhost", because the gateway for starknet-devnet is set incorrectly, when trying to get the string from the node.json file (is expecting a "localhost" key that doesn't exists).

**Solution**

When host is 0.0.0.0 create the node.json file with the following content:

```
{"localhost": "http://127.0.0.1:5050/"}
```

sokratisvas:
Fixes #171

### DIFF
--- a/src/nile/core/node.py
+++ b/src/nile/core/node.py
@@ -13,13 +13,14 @@ def node(host="127.0.0.1", port=5050, seed=None, lite_mode=False):
         file = NODE_FILENAME
         if host == "127.0.0.1":
             network = "localhost"
-        elif host == "0.0.0.0":
-            network = "localhost"
-            host = "127.0.0.1"
         else:
             network = host
-        gateway_url = f"http://{host}:{port}/"
-        gateway = {network: gateway_url}
+
+        if host == "0.0.0.0":
+            gateway = {"localhost": "http://127.0.0.1:5050/"}
+        else:
+            gateway_url = f"http://{host}:{port}/"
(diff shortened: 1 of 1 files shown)


ericnordelo (commented):
This fix won't allow to bind the server to all interfaces because the host in the call to starknet-devnet would be fixed to 127.0.0.1

The solution needs to write 127.0.0.1 to the node.json file while passing 0.0.0.0 to the cli call.

sokratisvas:
Thanks for the help! I think it is ok now.

ericnordelo (commented):
We still need to be able to set a different port when binding to all interfaces

martriay:
@sokratisvas any news on this?

Issue #180, opened 2022-09-02
Update Signer and artifacts to use the new Account API


martriay:
Migrate to camelCase, following https://github.com/OpenZeppelin/cairo-contracts/issues/382.

martriay:
Fixed by #184

PR #181, opened 2022-09-02
migrate account artifacts and signer to camelCase

martriay:
Fixes https://github.com/OpenZeppelin/nile/issues/180.


src/nile/artifacts/abis/Account.json
```diff
@@ -150,4 +150,4 @@
         ],
         "type": "function"
     }
-]
+]
```
andrew-fleming, line 153:
```suggestion
]

```
Hmm I don't remember seeing the "no newline" error in the past here

andrew-fleming (approved):
Very nice! Left a minor newline suggestion. Otherwise, this looks good to go :)

martriay:
supeceded by #184

Issue #188, opened 2022-09-11
Add working tests to the base project


martriay:
The current sample test is outdated and it shouldn't work with Cairo 0.10

PR #190, opened 2022-09-15
Update invoke to execute

ericnordelo:
Fixing the NotImplementedError raised by using invoke in the StarknetContractFunctionInvocation by using execute instead, as the cairo-lang repo recommends.


andrew-fleming (approved):

martriay:
This is tied to #188. Does it fix it fully?

ericnordelo:
> This is tied to #188. Does it fix it fully?

Yes, now the sample test works for version 0.10

martriay (approved):

PR #192, opened 2022-09-15
Devnet predeployed accounts

ericnordelo:
Add `get_predeployed_accounts` method and logic to `cli` and `nre` (compatible with starknet-devnet).

### DIFF
--- a/src/nile/cli.py
+++ b/src/nile/cli.py
@@ -19,6 +19,8 @@ from nile.core.test import test as test_command
 from nile.core.version import version as version_command
 from nile.utils.debug import debug as debug_command
 from nile.utils.get_accounts import get_accounts as get_accounts_command
+from nile.devnet import get_predeployed_accounts as get_predeployed_accounts_command
+
 
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
 
@@ -234,9 +236,13 @@ def debug(tx_hash, network, contracts_file):
 
 @cli.command()
 @network_option
-def get_accounts(network):
+@click.option("--predeployed/--registered", default=False)
+def get_accounts(network, predeployed):
     """Retrieve and manage deployed accounts."""
-    return get_accounts_command(network)
+    if not predeployed:
+        return get_accounts_command(network)
+    else:
+        return get_predeployed_accounts_command(network)
 
 
 cli = load_plugins(cli)
--- a/src/nile/cli.py
+++ b/src/nile/cli.py
@@ -17,10 +17,9 @@ from nile.core.plugins import load_plugins
 from nile.core.run import run as run_command
 from nile.core.test import test as test_command
 from nile.core.version import version as version_command
+from nile.devnet import get_predeployed_accounts as get_predeployed_accounts_command
 from nile.utils.debug import debug as debug_command
 from nile.utils.get_accounts import get_accounts as get_accounts_command
-from nile.devnet import get_predeployed_accounts as get_predeployed_accounts_command
-
 
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
 
--- a/src/nile/common.py
+++ b/src/nile/common.py
@@ -15,7 +15,7 @@ NODE_FILENAME = "node.json"
 RETRY_AFTER_SECONDS = 30
 
 
-def _get_gateway():
+def get_gateway():
     """Get the StarkNet node details."""
     try:
         with open(NODE_FILENAME, "r") as f:
@@ -27,7 +27,7 @@ def _get_gateway():
             f.write('{"localhost": "http://127.0.0.1:5050/"}')
 
 
-GATEWAYS = _get_gateway()
+GATEWAYS = get_gateway()
 
 
 def get_all_contracts(ext=None, directory=None):
--- a/src/nile/core/account.py
+++ b/src/nile/core/account.py
@@ -28,7 +28,7 @@ class Account:
                 self.alias = signer
             else:
                 self.signer = Signer(int(signer, 16))
-                self.alias = predeployed_info['alias']
+                self.alias = predeployed_info["alias"]
 
             self.network = network
         except KeyError:
@@ -40,8 +40,8 @@ class Account:
             return
 
         if predeployed_info is not None:
-            self.address = predeployed_info['address']
-            self.index = predeployed_info['index']
+            self.address = predeployed_info["address"]
+            self.index = predeployed_info["index"]
         elif accounts.exists(str(self.signer.public_key), network):
             signer_data = next(accounts.load(str(self.signer.public_key), network))
             self.address = signer_data["address"]
--- a/src/nile/core/account.py
+++ b/src/nile/core/account.py
@@ -20,11 +20,16 @@ load_dotenv()
 class Account:
     """Account contract abstraction."""
 
-    def __init__(self, signer, network):
+    def __init__(self, signer, network, predeployed_info=None):
         """Get or deploy an Account contract for the given private key."""
         try:
-            self.signer = Signer(int(os.environ[signer]))
-            self.alias = signer
+            if predeployed_info is None:
+                self.signer = Signer(int(os.environ[signer]))
+                self.alias = signer
+            else:
+                self.signer = Signer(int(signer, 16))
+                self.alias = predeployed_info['alias']
+
             self.network = network
         except KeyError:
             logging.error(
@@ -34,7 +39,10 @@ class Account:
             )
             return
 
-        if accounts.exists(str(self.signer.public_key), network):
+        if predeployed_info is not None:
+            self.address = predeployed_info['address']
+            self.index = predeployed_info['index']
+        elif accounts.exists(str(self.signer.public_key), network):
             signer_data = next(accounts.load(str(self.signer.public_key), network))
             self.address = signer_data["address"]
             self.index = signer_data["index"]
--- a/src/nile/devnet.py
+++ b/src/nile/devnet.py
@@ -1,11 +1,10 @@
-"""Starknet Devnet integration helpers"""
-import json
-import requests
+"""Starknet Devnet integration helpers."""
 import logging
 
-from nile.core.account import Account
-from nile.common import get_gateway
+import requests
 
+from nile.common import get_gateway
+from nile.core.account import Account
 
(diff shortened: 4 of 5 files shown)


src/nile/core/account.py
```diff
@@ -65,7 +78,10 @@ def deploy(self):
 
     def send(self, to, method, calldata, max_fee, nonce=None):
         """Execute a tx going through an Account contract."""
-        target_address, _ = next(deployments.load(to, self.network)) or to
+        target_address, _ = next(deployments.load(hex_address(to), self.network)) or to
+        # Work with integers internally
+        target_address = normalize_number(target_address)
```
andrew-fleming, line 83:
```suggestion
        target_address = hex(target_address)
```
`normalize_number` gave me a lot of trouble here. I think using `hex` is the way to go

src/nile/core/account.py
```diff
@@ -84,7 +100,7 @@ def send(self, to, method, calldata, max_fee, nonce=None):
         )
 
         return call_or_invoke(
-            contract=self.address,
+            contract=self,
```
andrew-fleming, line 103: This threw me off for a bit 😅

src/nile/core/call_or_invoke.py
```diff
@@ -5,13 +5,18 @@
...
 def call_or_invoke(
     contract, type, method, params, network, signature=None, max_fee=None
 ):
     """Call or invoke functions of StarkNet smart contracts."""
-    address, abi = next(deployments.load(contract, network))
+    if isinstance(contract, account.Account):
+        address = hex(contract.address)
+        abi = contract.abi_path
+    else:
+        address, abi = next(deployments.load(contract, network))
```
andrew-fleming, line 19:
```suggestion
    else:
        address, abi = next(deployments.load(contract, network))
        address = hex(address)
```
I believe the `address` should also be converted to `hex`. When I try to script a basic `get_balance` query, I'm receiving a type error:

`TypeError: expected str, bytes or os.PathLike object, not int`

src/nile/utils/__init__.py
```diff
@@ -116,3 +116,25 @@ def assert_event_emitted(tx_exec_info, from_address, name, data):
...
+def normalize_number(number):
+    """Normalize hex or int to int."""
+    if type(number) == str and number.startswith("0x"):
+        return int(number, 16)
+    else:
+        return int(number)
+
+
+def hex_address(number):
+    """Normalize hex or int to int."""
```
andrew-fleming, line 130: `hex_address` is trying to be a normalized number 🥸

tests/test_signer.py
```diff
@@ -35,7 +35,6 @@ async def send_transactions(signer, account, calls, nonce=None, max_fee=0):
     build_calls = []
     for call in calls:
         build_call = list(call)
-        build_call[0] = hex(build_call[0])
         build_calls.append(build_call)
```
andrew-fleming, line 38: Since we don't need to hexify the first arg, we can remove all this and just pass `calls` to `get_raw_invoke` and `sign_transaction`

Also, let's remember to update `MockSigner` and `MockEthSigner` in Contracts before the next release since they use the same `build_calls` logic

andrew-fleming (commented):
Great work, @ericnordelo! I left some comments and questions :)

README.md
```diff
@@ -388,7 +388,34 @@ def run(nre):
...
+
+### `get-accounts --predeployed`
+
+This flag retrieves the predeployed accounts if the network you are connecting to is a [starknet-devnet](https://github.com/Shard-Labs/starknet-devnet) instance.
+
+You can use it either from the cli:
+
+```sh
+nile get-accounts --predeployed
+```
```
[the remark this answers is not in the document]
ericnordelo: We can't now, but it is an interesting idea. Maybe it will be worth modifying the **send** command (with a predeployed flag or similar) to allow sending tx through predeployed accounts? I'm curious about both your opinions on this @andrew-fleming @martriay

src/nile/cli.py
```diff
@@ -234,9 +237,13 @@ def debug(tx_hash, network, contracts_file):
 
 @cli.command()
 @network_option
-def get_accounts(network):
+@click.option("--predeployed/--registered", default=False)
```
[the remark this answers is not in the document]
ericnordelo: I was checking Click, and as you can check [here](https://click.palletsprojects.com/en/8.1.x/options/#boolean-flags), Click always wants you to provide an enable and disable flag so that you can change the default later, but just for this, the disable flag is not necessary and is not required from CLI, so I thought that with documenting the --predeployed is enough, because is never needed to pass the --registered flag. Let me know what you think.

src/nile/common.py
```diff
@@ -74,6 +76,10 @@ def run_command(
 
 def get_nonce(contract_address, network):
     """Get the current nonce for contract address in a given network."""
+    # Starknet CLI requires an hex string for get nonce command
+    if not str(contract_address).startswith("0x"):
+        contract_address = hex(int(contract_address))
```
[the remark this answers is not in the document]
ericnordelo: Yes, and also I just noticed that this method shouldn't receive anything not being an int in contract address (the design patter for hex and int we are trying to follow), so this parser should be refactored accordingly too.

src/nile/core/account.py
```diff
@@ -20,11 +21,16 @@
 class Account:
     """Account contract abstraction."""
 
-    def __init__(self, signer, network):
+    def __init__(self, signer, network, predeployed_info=None):
```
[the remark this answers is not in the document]
ericnordelo: Agree, I've just created it.

src/nile/core/account.py
```diff
@@ -65,7 +78,10 @@ def deploy(self):
 
     def send(self, to, method, calldata, max_fee, nonce=None):
         """Execute a tx going through an Account contract."""
-        target_address, _ = next(deployments.load(to, self.network)) or to
+        target_address, _ = next(deployments.load(hex_address(to), self.network)) or to
```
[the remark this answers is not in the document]
ericnordelo: This was not compliant with the pattern from my side because all the methods should work with int. The solution I used is just to remove the hex_address here and add the normalize_number to the address as input in deployments.load. This allows passing to as an alias (I didn't notice this, docstring for params would be super helpful as you suggested).

src/nile/core/account.py
ericnordelo: What troubles did you find? Using hex here is against the pattern we agreed, we should use int internally. Anyway, normalize_number is not necessary here because deployments.load always returns an int (being compliant with the pattern). I removed it.

src/nile/core/account.py
ericnordelo: I think is better to pass the account instance to avoid passing a lot of extra parameters like the ABI (that was required for predeployed accounts). All these params are in the instance already. Suggestions are welcome to address this, of course.

src/nile/core/call_or_invoke.py
ericnordelo: I added it outside the if/else for both branches before calling the starknet cli.

tests/test_signer.py
ericnordelo: Fixed test.

README.md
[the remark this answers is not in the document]
andrew-fleming: I assumed as much, just confirming. What do you think about removing the CLI snippet and the `You can use it either from the cli`? I don't think there's any value from running the command in the CLI and it might avoid users (or me) asking if we can send txs from the predeployeds through the CLI

> Maybe it will be worth modifying the send command (with a predeployed flag or similar) to allow sending tx through predeployed accounts

Hmm it sounds like it'd be great to consider. I wouldn't include it in this PR though, that seems like a down-the-road feature

src/nile/cli.py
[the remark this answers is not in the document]
andrew-fleming: Yeah, that's reasonable. Documenting `--predeployed` seems like enough

README.md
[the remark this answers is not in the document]
ericnordelo: But you can use it from the cli, just not to send transactions. You can list the predeployed accounts from the cli, as you can do with registered accounts running `nile get-accounts`, so I think thee snippet still makes sense.

src/nile/core/call_or_invoke.py
andrew-fleming: Perfect!

src/nile/core/account.py
andrew-fleming: Agreed on this being a better approach. Well done!

src/nile/core/account.py
andrew-fleming: I believe the issues were from how I was hacking the `deployments.load` when trying to get it to work locally. With the latest updates to the PR, everything's running well

src/nile/core/account.py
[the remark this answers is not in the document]
andrew-fleming: Excellent! Yeah, I just wanted to throw out a workaround solution to potentially aid in the adjustment. Works well now!

andrew-fleming (dismissed):
Excellent improvements! Looks good to go

src/nile/core/account.py
martriay: this is causing a circular dependency, i'm rolling it back in #268 

```sh
____________________________________________________ ERROR collecting tests/commands/test_get_balance.py ____________________________________________________
ImportError while importing test module '/home/nile/tests/commands/test_get_balance.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/commands/test_get_balance.py:7: in <module>
    from nile.utils.get_balance import get_balance
.tox/default/lib/python3.8/site-packages/nile/utils/get_balance.py:3: in <module>
.tox/default/lib/python3.8/site-packages/nile/core/call_or_invoke.py:6: in <module>
    from nile.core import account
.tox/default/lib/python3.8/site-packages/nile/core/account.py:20: in <module>
E   ImportError: cannot import name 'call_or_invoke' from partially initialized module 'nile.core.call_or_invoke' (most likely due to a circular import) (/home/nile/.tox/default/lib/python3.8/site-packages/nile/core/call_or_invoke.py)
```

PR #197, opened 2022-09-19
Add get-nonce to CLI and NRE

ericnordelo:


andrew-fleming (commented):
Looking good, @ericnordelo! I left a couple questions and comments

src/nile/utils/get_nonce.py
```diff
@@ -0,0 +1,25 @@
...
+    if network == "mainnet":
+        os.environ["STARKNET_NETWORK"] = "alpha-mainnet"
+    elif network == "goerli":
+        os.environ["STARKNET_NETWORK"] = "alpha-goerli"
+    else:
+        command.append(f"--feeder_gateway_url={GATEWAYS.get(network)}")
+
+    nonce = int(subprocess.check_output(command).strip())
+
+    logging.info(f"\nCurrent nonce for {contract_address} is {nonce}")
```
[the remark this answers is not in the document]
ericnordelo: Moving logging to CLI would break the consistency with the other commands logging from the implementation and not from the CLI. To fix the issue, I separated the log from the logic in the get-nonce.py file, and I'm using the logic without logging in account.send, let me know what you think.

src/nile/utils/get_nonce.py
```diff
@@ -0,0 +1,25 @@
+"""Retrieve and manage deployed accounts."""
```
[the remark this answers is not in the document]
ericnordelo: :(

tests/commands/test_get_nonce.py
```diff
@@ -0,0 +1,41 @@
...
+        command.append("--feeder_gateway_url=http://127.0.0.1:5050/")
+    else:
+        assert os.getenv("STARKNET_NETWORK") == f"alpha-{network}"
+
+    mock_subprocess.assert_called_once_with(command)
+
+
+@patch("nile.core.node.subprocess.check_output", return_value="5")
+@patch("nile.common.get_gateway", return_value=GATEWAYS)
+def test_deploy(mock_gateway, mock_subprocess, caplog):
```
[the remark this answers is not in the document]
ericnordelo: Fixed.

src/nile/utils/get_nonce.py
```diff
@@ -0,0 +1,25 @@
...
+import subprocess
+
+from nile.common import get_gateway
+
+GATEWAYS = get_gateway()
+
+
+def get_nonce(contract_address, network):
+    """Get the current nonce for contract address in a given network."""
+    command = ["starknet", "get_nonce", "--contract_address", contract_address]
```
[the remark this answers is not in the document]
ericnordelo: Yes, I added the feature by parsing ints to hexs. This time is Starknet CLI the one not accepting int, but only hex strings. This breaks our idea of only converting to int from the user input and working with int internally. We were assuming that Starknet allowed int for every interaction and here we have an exception.

src/nile/utils/get_nonce.py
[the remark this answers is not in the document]
andrew-fleming: Ahh yes, I see the consistency issue. This solution looks good to me!

src/nile/utils/get_nonce.py
```diff
@@ -0,0 +1,35 @@
...
+    logging.info(f"Current Nonce: {nonce}")
+
+    return nonce
+
+
+def get_nonce_without_log(contract_address, network):
+    """Get the current nonce without logging."""
+    # Starknet CLI requires an hex string for get nonce command
+    if type(contract_address) == int:
+        contract_address = hex(contract_address)
```
andrew-fleming, line 24:
```suggestion
    if not contract_address.startswith("0x"):
        contract_address = hex(int(contract_address))
```

After playing around with this, it seems `contract_address` is always passed as a string. The above suggestion is how I got it to work. Let me know what you think

ericnordelo: In case the method for some reason receives a int, I ended up with this from your suggestion:

```
if not str(contract_address).startswith("0x"):
    contract_address = hex(contract_address)
```

I will add a test specifically for this.

andrew-fleming: Since `contract_address` is a string, I believe we'll have to convert it to an `int` before `hex`

> I will add a test specifically for this.

Great idea!

ericnordelo: Yes indeed, I will push with the update and the test.

src/nile/utils/get_nonce.py
```diff
@@ -0,0 +1,35 @@
...
+from nile.common import get_gateway
+
+GATEWAYS = get_gateway()
+
+
+def get_nonce(contract_address, network):
+    """Get the current nonce for contract address in a given network."""
+    nonce = get_nonce_without_log(contract_address, network)
+
+    logging.info(f"Current Nonce: {nonce}")
```
martriay, line 15: i wonder whether we should print this here or directly on the CLI, for consistency with other commands. the only counter argument i can think of is that i've been thinking that having the ability to turn on/off nile's logging from nre could be useful, i'd lean towards opt-in rather than opt-out logging though, but not sure.

tests/commands/test_get_nonce.py
```diff
@@ -0,0 +1,41 @@
...
+        command.append("--feeder_gateway_url=http://127.0.0.1:5050/")
+    else:
+        assert os.getenv("STARKNET_NETWORK") == f"alpha-{network}"
+
+    mock_subprocess.assert_called_once_with(command)
+
+
+@patch("nile.core.node.subprocess.check_output", return_value="5")
+@patch("nile.common.get_gateway", return_value=GATEWAYS)
+def test_get_nonce(mock_gateway, mock_subprocess, caplog):
```
martriay, line 33: depending on what's decided re: my above comment, this could be moved to the cli's suite

martriay (commented):
looking good! added a question/comment on top of the existing ones

src/nile/utils/get_nonce.py
andrew-fleming: I should mention, Nile's `logging` doesn't print for the NRE (I assume it's lost in the abstraction, but I'd have to look deeper to give a better answer). If we want the output from these functions, I know `print` returns the output through the NRE. I'm sure there's a better way with `logging` though (if indeed we want this)

ericnordelo: Currently, other commands are printing from the command code, not from the CLI, so I added the logging here for consistency.

To disable logging from nre, right now is as easy as running `logging.disable(logging.INFO)` right before executing the call to the command. This is opt-out, but we could add opt-in by adding a param in each nre command default to True,  disabling the logging like this (ERROR messages would still be logged, but maybe this is a expected behavior, if not we can disable all logging using `logging.disable(logging.CRITICAL)`).

tests/commands/test_get_nonce.py
ericnordelo: You mean the logging assertion, not the full test, right? The result assertion still makes sense in that case, I think.

src/nile/utils/get_nonce.py
martriay: that's true i got confused by the fact that nre commands today don't print by default, but it should be handled from logging not print. so i guess it's a matter of turning it on? that'd be the opt-in way i was thinking about, and i was thinking for it to be global for all commands, although an interface to do it per-command is not a bad idea.

the only doubt that remains to me is whether we should make this turning on/off logging independent of logging's global config, to not interfere with the user's script in case they use the module too directly or through other library.

ericnordelo: I think we can get the specific logger, and disable from there instead of the global logging object, similar to what we are doing in run.py for scripting. Not sure if we should name this logger (I would need to go a little bit deeper in the logging library), but I think we can achieve what we want like this.

ericnordelo: Also, I noticed some print statements in some commands, should we refactor those to logging too?

andrew-fleming: Similar to @ericnordelo's solution, we could just include a `verbose` flag in `run`. I think this will follow the opt-in approach without too much refactoring i.e. `nile run scripts/deploy.py --verbose`.

Something like this:

```
def run(path, network, verbose):
    """Run nile scripts passing on the NRE object."""
    logger = logging.getLogger()
    if not verbose:
        logger.disabled = True
    script = SourceFileLoader("script", path).load_module()
    nre = NileRuntimeEnvironment(network)
    script.run(nre)
```

Further, if users use another lib or use the modules too closely, they can set up their own log handling. I think keeping it in for outside cases is the more flexible approach.

ericnordelo: I agree, but I think it would be helpful to name the logger so that the user can silence or activate logs specifically for the package from other integrations, I'm not sure how this works, but to silence logs from urllib3 requests, I can run `logging.getLogger("urllib3").setLevel(logging.WARNING)`, we can do something similar.

ericnordelo: I created a separated issue for this.

ericnordelo:
@andrew-fleming @martriay , Besides the get-nonce feature, this PR adds some logic relative to the int/hex pattern (on top of the predeployed accounts PR).

tox.ini
```diff
@@ -15,7 +15,7 @@ extras =
     testing
 deps =
     cairo-lang==0.10.0
-    starknet-devnet==0.3.0
+    starknet-devnet==0.3.1
```
andrew-fleming, line 18: I'm thinking we should probably set `starknet-devnet` in the `setup.cfg` reqs as well. There's already an issue for tackling dep management (#134). As it stands now, I believe users have to manually install the devnet. Just leaving this as a reminder

andrew-fleming (commented):
Looking good, sir! Left some small suggestions and comments

tox.ini
ericnordelo: Right now `nile init` installs devnet (without specifying version), potentially bringing issues. For example 0.3.0 predeployed accounts doesn't work with the current Account interface, totally agree with you, we should handle these requirements.

src/nile/deployments.py
```diff
@@ -49,9 +49,15 @@ def register_class_hash(hash, network, alias):
...
+def exists(address_or_alias, network):
+    """
+    Return whether a deployment exists or not.
+
+    If address_or_alias is an int, address is assumed.
+
+    If address_or_alias is a str, alias is assumed.
+    """
+    foo = next(load(address_or_alias, network), None)
     return foo is not None
```
andrew-fleming, line 61:
```suggestion
    deployment = next(load(address_or_alias, network), None)
    return deployment is not None
```
Don't forget about `foo`!

src/nile/cli.py
```diff
@@ -115,15 +118,17 @@ def setup(signer, network):
 @click.argument("params", nargs=-1)
 @click.option("--max_fee", nargs=1)
 @network_option
-def send(signer, contract_name, method, params, network, max_fee=None):
+def send(signer, address_or_alias, method, params, network, max_fee=None):
```
andrew-fleming, line 121: Good call! I believe we also have to update the `click.argument` to `address_or_alias`

andrew-fleming (commented):
Good additions! Left a comment on `send`. We should update the `click.argument` for `invoke` and `call` also :)

src/nile/cli.py
ericnordelo: Yes, I missed that. Updated already :)

andrew-fleming (approved):
Looks good to go 🚀

src/nile/core/declare.py
```diff
@@ -15,8 +15,10 @@ def declare(contract_name, network, alias=None, overriding_path=None):
...
     output = run_command(contract_name, network, overriding_path, operation="declare")
     class_hash, tx_hash = parse_information(output)
-    logging.info(f"⏳ Declaration of {contract_name} successfully sent at {class_hash}")
-    logging.info(f"🧾 Transaction hash: {tx_hash}")
+    logging.info(
+        f"⏳ Declaration of {contract_name} successfully sent at {hex(class_hash)}"
+    )
+    logging.info(f"🧾 Transaction hash: {hex(tx_hash)}")
 
     deployments.register_class_hash(class_hash, network, alias)
```
martriay, line 23: why isn't the class hash stored as hex too? it should be considered external to me, since it's available to users/plugins/external tooling

ericnordelo: The class hash is stored as hex (is being converted in the `register_class_hash` following the logging lines).

It is not converted here and passed as an argument because that is against the pattern (the current definition, at least), which states:

- Accepting and using only integers for Internal APIs (register_class_hash is an internal API)
- Convert to hex before writing to files (we are doing that inside the method)

This is the current definition of the pattern: https://github.com/OpenZeppelin/nile/issues/216

Issue #203, opened 2022-09-22
Add option to `deploy` with a different ABI 


ericglau:
For `deploy`, the ABI that gets registered in the deployments file is currently the ABI associated with the provided contract.

When deploying a proxy contract, I want to register the ABI of the implementation, not of the proxy, so that calling `invoke` or `call` on that address later will use the implementation's ABI.

PR #206, opened 2022-09-22
Add abi parameter to `deploy`

ericglau:
Fixes https://github.com/OpenZeppelin/nile/issues/203


andrew-fleming (commented):
@ericglau excellent addition! I think we just need to update the docs with new param as well

ericglau:
@andrew-fleming The current docs are mostly for the CLI, but this new parameter is only for the NRE's `deploy` function and not exposed in the CLI.  Would it be useful to add this parameter to the CLI as well?

andrew-fleming:
Great question. Maybe we should include it in the CLI just for consistency with documenting the param? Otherwise, I think it'll be difficult for users to know about this feature. What do you think?

martriay:
+1 to add it to the CLI

ericglau:
Added the param to cli and docs.

andrew-fleming (approved):
Looks good to me!

Issue #205, opened 2022-09-22
Improve how `declare` handles an existing class hash


ericglau:
When I call `declare` on an implementation contract to be used with a proxy, if it was previously declared, `register_class_hash` currently raises an exception containing the hash in the message string.  

To improve usability, I want a way to get the existing declaration's class hash (without having to parse the error message) so that it can be used afterwards.

Consider one of the following:
~~1. Use a custom error that contains the hash in a field, so that the caller of `declare` can catch the error and retrieve the hash
2. In `declare`, suppress the error if a class hash already exists, and just proceed to return with the existing class hash
3. In `declare`, suppress the error if a class hash already exists, and update the entry in the declaration file (e.g. if a different alias is provided)
4. In `declare`, have an optional flag such as `update` which allows it to do (3) only if the flag is set.~~
5. Provide a function similar to `get_declaration`, but takes a contract (rather than an alias) and returns the corresponding class hash.

ericglau:
Alternatively, should `declare` suppress the error if a class hash already exists, and/or should it simply update the entry (e.g. if a different alias is provided)?

ericglau:
I prefer option (4) above, since an `update` flag would make `deploy` easier to use in this scenario, and still retain the original error if an update is not desired.

martriay:
the equivalent of (5) would be:

```python
from starkware.crypto.signature.fast_pedersen_hash import pedersen_hash
from starkware.starknet.core.os.class_hash import compute_class_hash

def get_hash(contract_class):
    return compute_class_hash(
        contract_class=contract_class,
        hash_func=pedersen_hash
    )
```

Issue #195, opened 2022-09-19
Refactor project for better code organization


ericnordelo:

ericnordelo:
I left the proposed specification in the linked PR.

PR #207, opened 2022-09-22
Add Account.declare

### LINKED ISSUE #186: Adjust `declare` so only accounts can declare txs
As mentioned in the release notes: [https://github.com/starkware-libs/cairo-lang/releases](https://github.com/starkware-libs/cairo-lang/releases)

martriay:
Fixes #186, #205.

- This PR also refactors `run_command` so it actually centralizes starknet CLI calls.
- It also implements `nile.core.declare.get_hash` which takes a `contract_class` and returns its hash (see #205).
- An upcoming problem with this PR is that it also enables `Account.deploy_contract` which relies on a [Deployer Contract](https://github.com/OpenZeppelin/cairo-contracts/pull/467/) to deploy, which [eludes our deployments tracking/management system](https://github.com/OpenZeppelin/nile/issues/186#issuecomment-1255581069). I think we need to rethink its model, but probably in a different issue/PR.


src/nile/core/declare.py
```diff
@@ -26,3 +49,8 @@ def alias_exists(alias, network):
     """Return whether an alias exists or not."""
     existing_alias = next(deployments.load_class(alias, network), None)
     return existing_alias is not None
+
+
+def get_hash(contract_class):
```
ericglau, line 54: Is contract class a concept that users are expected to use / be familiar with?  Would it be simpler if this takes a `contract_name` (and perhaps an `overriding_path`) to be consistent with `declare`, and have this function load the artifact instead of requiring the user to do so?

andrew-fleming (commented):
Excellent PR! I left some questions and comments :)

src/nile/core/declare.py
andrew-fleming: The `contract_name` approach seems like a good idea^.

Also, it's a little unclear to me how this function should be used. I assume the intent as-is requires a direct import. If so, would `utils` be a better spot? Or should we add this to the NRE?

martriay: i'll move it to utils then. usage is described in https://github.com/OpenZeppelin/nile/issues/205. I don't think utils in general would make good commands, although it depends on usage/demand.

src/nile/cli.py
```diff
@@ -88,12 +87,23 @@ def deploy(artifact, arguments, network, alias):
...
 
 @cli.command()
-@click.argument("artifact", nargs=1)
+@click.argument("signer", nargs=1)
+@click.argument("contract_name", nargs=1)
+@click.option("--max_fee", nargs=1)
 @network_option
 @click.option("--alias")
-def declare(artifact, network, alias):
+@click.option("--directory")
```
andrew-fleming, line 95:
```suggestion
@click.option("--contracts_directory")
```

andrew-fleming (approved):
Small fix on the click option, but otherwise this looks good to me!

src/nile/cli.py
martriay: chose an alternative version. i prefer keeping the cli arg name short, what do you think?

andrew-fleming: Easier to type, totally fine by me

src/nile/utils/__init__.py
```diff
@@ -3,8 +3,13 @@
 import math
 from pathlib import Path
 
+from nile.common import ABIS_DIRECTORY, BUILD_DIRECTORY
+
 try:
+    from starkware.crypto.signature.fast_pedersen_hash import pedersen_hash
     from starkware.starknet.business_logic.execution.objects import Event
+    from starkware.starknet.compiler.compile import compile_starknet_files
```
ericglau, line 11:
```suggestion
    from starkware.starknet.services.api.contract_class import ContractClass
```

src/nile/utils/__init__.py
```diff
@@ -116,3 +121,14 @@ def assert_event_emitted(tx_exec_info, from_address, name, data):
...
+
+def get_hash(contract_name, overriding_path=None):
+    """Return the class_hash of a given contract class."""
+    base_path = (
+        overriding_path if overriding_path else (BUILD_DIRECTORY, ABIS_DIRECTORY)
+    )
+    contract_class = compile_starknet_files(
+        files=[f"{base_path[1]}/{contract_name}.cairo"], debug_info=True
+    )
+    return compute_class_hash(contract_class=contract_class, hash_func=pedersen_hash)
```
ericglau, line 134: - The .cairo file shouldn't be based on the base_path which is for artifacts.  I don't think we need to compile here, but can expect that contracts have already been compiled (other functions such as `deploy` also expect this).  So we can just read the artifact json.
- For convenience, this should return hex (`compute_class_hash` returns an integer).  The deployments file also uses hex.

```suggestion
def get_hash(contract_name, overriding_path=None):
    """Return the class hash (in hex) of a given contract class."""
    base_path = (
        overriding_path if overriding_path else (BUILD_DIRECTORY, ABIS_DIRECTORY)
    )
    with open(f"{base_path[0]}/{contract_name}.json", "r") as fp:
        contract_class = ContractClass.loads(fp.read())

    return hex(compute_class_hash(contract_class=contract_class, hash_func=pedersen_hash))
```

README.md
```diff
@@ -188,6 +190,26 @@ Some things to note:
 - `max_fee` defaults to `0`. Add `--max_fee <max_fee>` to set the maximum fee for the transaction
 - `network` defaults to the `localhost`. Add `--network <network>` to change the network for the transaction
 
+### `declare`
```
ericglau, line 193: The old `declare` doc section still exists on line 129 - should that be removed?

src/nile/core/account.py
```diff
@@ -63,9 +65,59 @@ def deploy(self):
...
+    ):
+        """Deploy a contract through an Account contract."""
+        return self.send(
+            to=deployer_address or UNIVERSAL_DEPLOYER_ADDRESS,
+            method="deployContract",
+            calldata=[class_hash, salt, unique, len(calldata), *calldata],
+            max_fee=max_fee,
+        )
+
     def send(self, to, method, calldata, max_fee, nonce=None):
```
ericglau, line 114:
```suggestion
    def send(self, to, method, calldata, max_fee=None, nonce=None):
```

src/nile/core/account.py
```diff
@@ -63,9 +65,59 @@ def deploy(self):
 
         return address, index
 
+    def declare(
+        self, contract_name, max_fee, nonce=None, alias=None, contracts_directory=None
```
ericglau, line 69:
```suggestion
        self, contract_name, max_fee=None, nonce=None, alias=None, contracts_directory=None
```

src/nile/core/account.py
```diff
@@ -63,9 +65,59 @@ def deploy(self):
...
+            sender=self.address,
+            contract_name=contract_name,
+            signature=[sig_r, sig_s],
+            alias=alias,
+            network=self.network,
+            max_fee=max_fee,
+        )
+
+    def deploy_contract(
+        self, class_hash, salt, unique, calldata, max_fee, deployer_address=None
```
ericglau, line 104:
```suggestion
        self, class_hash, salt, unique, calldata, max_fee=None, deployer_address=None
```

ericglau (commented):
account's `declare`, `deploy_contract`, and `send` should set `max_fee` to be `None` by default

src/nile/core/account.py
```diff
@@ -63,9 +65,59 @@ def deploy(self):
...
+        if nonce is None:
+            nonce = get_nonce(self.address, self.network)
+
+        if max_fee is None:
+            max_fee = 0
+        else:
+            max_fee = int(max_fee)
+
+        contract_class = compile_starknet_files(
+            files=[f"{contracts_directory}/{contract_name}.cairo"], debug_info=True
```
ponderingdemocritus, line 84: do we need to compile in this location? It would be assumed that the contract being declared has been compiled before running this.

src/nile/core/account.py
```diff
@@ -3,10 +3,12 @@
 import os
 
 from dotenv import load_dotenv
+from starkware.starknet.compiler.compile import compile_starknet_files
 
 from nile import accounts, deployments
-from nile.common import get_nonce
+from nile.common import CONTRACTS_DIRECTORY, UNIVERSAL_DEPLOYER_ADDRESS, get_nonce
```
ericnordelo, line 9: `get_nonce` is part of `nile.utils` now (because is a command). And we should use get_nonce_without_log (as in the `send` method) to avoid logging nonce from the cli in the declare command.

Regarding where `get_nonce` should be (maybe in the commands folder inside core), this is related to this issue that is assigned to the **after** milestone: #195 .

ericnordelo (commented):
I left a couple of comments, but I can provide a better review after conflicts with main are resolved, because there are important differences regarding the int pattern mostly.

README.md
ericglau: I meant there are currently 2 `declare` sections.

src/nile/signer.py
```diff
@@ -68,6 +83,18 @@ def from_call_to_call_array(calls):
     return (call_array, calldata)
 
 
+def get_declare_hash(sender, contract_class, max_fee, nonce):
+    """Compute the hash of a declare transaction."""
+    return calculate_declare_transaction_hash(
+        contract_class=contract_class,
+        chain_id=StarknetChainId.TESTNET.value,
```
andrew-fleming, line 90: At some point, we're going to need a param to adjust the chain id based on the network (for `get_transaction_hash` as well). I just created an issue #234.

andrew-fleming (commented):
Looking good, @martriay! I left a few comments :)

ericglau:
The `get_hash` logic looks good to me, thanks! (I didn't review the rest of this PR)

src/nile/nre.py
```diff
@@ -25,9 +24,15 @@ def compile(self, contracts):
...
 
-    def declare(self, contract, alias=None, overriding_path=None):
+    def declare(self, signer, contract, alias=None, max_fee=None, overriding_path=None):
         """Declare a smart contract class."""
-        return declare(contract, self.network, alias)
+        account = Account(signer, self.network)
+        return account.declare(
+            contract,
+            alias=alias,
+            max_fee=max_fee,
```
andrew-fleming, line 33: Should we remove this from the NRE? We don't include `send` since it's already a method of `Account`. I'm thinking the same approach with `declare` should be fine i.e. `account.declare(...)`.

src/nile/core/account.py
```diff
@@ -77,16 +82,59 @@ def deploy(self):
 
         return address, index
 
-    def send(self, address_or_alias, method, calldata, max_fee, nonce=None):
-        """Execute a tx going through an Account contract."""
-        if not is_alias(address_or_alias):
-            address_or_alias = normalize_number(address_or_alias)
+    def declare(
```
ericnordelo, line 85: In this release, we are adding query calls to the `invoke` command. We should add the query flags to the `declare` command too (**estimate_fee** and **simulate**). Even if it is not part of this PR, I think it would be good to include it, at least in the next release.

ericnordelo (commented):
Looking good! I added a comment and thumbs up to the Andrew review notes.

src/nile/core/account.py
martriay: created https://github.com/OpenZeppelin/nile/issues/242

ericnordelo (dismissed):
Looking good to go!

andrew-fleming (dismissed):
Looks good to go!

PR #210, opened 2022-09-23
Add short strings from cli

ericnordelo:
Update `prepare_params` method in common to process short strings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,7 +421,7 @@ def run(nre):
 
 From [cairo-lang docs](https://www.cairo-lang.org/docs/how_cairo_works/consts.html#short-string-literals): A short string is a string whose length is at most 31 characters, and therefore can fit into a single field element.
 
-In Nile, arguments to contract calls (calldata) that are neither int nor hex, are treated as short strings and converted automatically to the corresponding felt representation. Because of this, you can run the next from the CLI:
+In Nile, arguments to contract calls (calldata) that are neither int nor hex, are treated as short strings and converted automatically to the corresponding felt representation. Because of this, you can run the following from the CLI:
 
 ```sh
 nile deploy MyToken 'MyToken name' 'MyToken symbol' (...)
--- a/README.md
+++ b/README.md
@@ -433,7 +433,14 @@ And this is equivalent to passing the felt representation directly like this:
 nile deploy MyToken 0x4d79546f6b656e206e616d65 0x4d79546f6b656e2073796d626f6c (...)
 ```
 
-Note that if you want to pass the token name as a hex or an int, you need to provide the felt representation directly because these values are not interpreted as short strings.
+Note that if you want to pass the token name as a hex or an int, you need to provide the felt representation directly because these values are not interpreted as short strings. You can open a python terminal, and import and use the `str_to_felt` util like this:
+
+```python
+>>> from nile.utils import str_to_felt
+>>>
+>>> str_to_felt('any string')
+460107418789485453340263
+```
 
 ## Extending Nile with plugins
 
--- a/README.md
+++ b/README.md
@@ -390,6 +390,24 @@ def run(nre):
 
 > Please note that the list of accounts include only those that exist in the local `<network>.accounts.json` file.
 
+## Short string literals
+
+From [cairo-lang docs](https://www.cairo-lang.org/docs/how_cairo_works/consts.html#short-string-literals): A short string is a string whose length is at most 31 characters, and therefore can fit into a single field element.
+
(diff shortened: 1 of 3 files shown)


src/nile/common.py
```diff
@@ -93,16 +95,38 @@ def parse_information(x):
...
+    try:
+        int(param)
+    except Exception:
+        is_int = False
+
+    # converting to hex
+    try:
+        int(param, 16)
+    except Exception:
+        is_hex = False
```
andrew-fleming, line 130: Should we also include the check `param.startswith("0x")`? Otherwise, I think shorts strings like "bad" will be interpreted as hex and not convert to a felt

andrew-fleming (commented):
> What if someone wants to pass a short string that is an in or hex? (like token name 0xA) In the current implementation this would be passed as hex, no as string.

Hmm my initial thought is to have the user manually convert the string to felt (or hex) for edge cases like you mentioned. If we're insistent on handling this internally, we could refactor how we handle calldata with click—maybe using a flag after a calldata arg that signifies it should be interpreted as a short string i.e. `nile send <KEY> <ADDRESS> 0xA --s`

I'm curious to hear other potential solutions on this.

> If user sends a string to large the call is going to revert with ValueError: Input value out of bounds, maybe a more user friendly response?

Yes, totally agree

src/nile/common.py
ericnordelo: Good catch! Agree.

ericnordelo:
I think that updating the docs to reflect that if you pass an int or a hex, those are going to be treated as felt (not as a short string) is enough, and a better UX than converting all the short strings to felt manually. With this solution user needs to convert manually to felt only if the short string itself is an int or a hex, otherwise, there shouldn't be further problems (this must be reflected in the docs).

README.md
```diff
@@ -417,6 +417,24 @@ def run(nre):
...
+nile deploy MyToken 'MyToken name' 'MyToken symbol' (...)
+```
+
+And this is equivalent to passing the felt representation directly like this:
+
+```sh
+nile deploy MyToken 0x4d79546f6b656e206e616d65 0x4d79546f6b656e2073796d626f6c (...)
+```
+
+Note that if you want to pass the token name as a hex or an int, you need to provide the felt representation directly because these values are not interpreted as short strings.
```
andrew-fleming, line 436: Do you think it'd be helpful to give a quick example on how to manually convert? Maybe just a link to `str_to_felt`?

andrew-fleming (commented):
Excellent work, @ericnordelo! I left a small question; otherwise, I think we're good to go!

README.md
andrew-fleming: Or perhaps this should be expected considering that's how it was prior. Thoughts?

ericnordelo: I think adding the how-to is helpful in any case, so I added an entry for this. Let me know what you think, and any updates to the text (my English is far from good) is appreciated.

README.md
```diff
@@ -417,6 +417,31 @@ def run(nre):
     bob.send(...)
 ```
 
+## Short string literals
+
+From [cairo-lang docs](https://www.cairo-lang.org/docs/how_cairo_works/consts.html#short-string-literals): A short string is a string whose length is at most 31 characters, and therefore can fit into a single field element.
+
+In Nile, arguments to contract calls (calldata) that are neither int nor hex, are treated as short strings and converted automatically to the corresponding felt representation. Because of this, you can run the next from the CLI:
```
andrew-fleming, line 424:
```suggestion
In Nile, arguments to contract calls (calldata) that are neither int nor hex, are treated as short strings and converted automatically to the corresponding felt representation. Because of this, you can run the following from the CLI:
```

README.md
```diff
@@ -417,6 +417,31 @@ def run(nre):
...
+nile deploy MyToken 0x4d79546f6b656e206e616d65 0x4d79546f6b656e2073796d626f6c (...)
+```
+
+Note that if you want to pass the token name as a hex or an int, you need to provide the felt representation directly because these values are not interpreted as short strings. You can open a python terminal, and import and use the `str_to_felt` util like this:
+
+```python
+>>> from nile.utils import str_to_felt
+>>>
+>>> str_to_felt('any string')
+460107418789485453340263
```
andrew-fleming, line 442: 👍

andrew-fleming (dismissed):
Looks great!

> any updates to the text (my English is far from good) is appreciated.

Haha your English seems quite good to me :) I left one tiny suggestion

README.md
ericnordelo: suggestion committed, ty

andrew-fleming (approved):

PR #211, opened 2022-09-23
Add function to update registered deployment ABI

### LINKED ISSUE #204: Add function to update a deployment in the deployments file
The deployments file can currently be appended only with new deployments.

For a proxy deployment, after the proxy has been upgraded, I want to update its corresponding address entry in the deployment file to use the new implementation's ABI.

ericglau:
Fixes #204

The `update_abi` function takes an identifier (address or alias) as input, and updates the ABI while keeping the same address and alias.

### DIFF
--- a/src/nile/deployments.py
+++ b/src/nile/deployments.py
@@ -32,19 +32,29 @@ def update(address, abi, network, alias):
     if not os.path.exists(file):
         raise Exception(f"{file} does not exist")
 
+    if alias is not None:
+        if exists(alias, network):
+            raise Exception(f"Alias {alias} already exists in {file}")
+
     with open(file, "r") as fp:
         lines = fp.readlines()
 
     found = False
     for i in range(len(lines)):
-        line_address = lines[i].strip().split(":")[0]
-        if line_address == address:
+        line = lines[i].strip().split(":")
+        current_address = line[0]
+
+        new_alias = alias
+        if new_alias is None and len(line) > 2:
+            new_alias = line[2]
+
+        if current_address == address:
             found = True
 
             replacement = f"{address}:{abi}"
-            if alias is not None:
-                logging.info(f"📦 Updating deployment as {alias} in {file}")
-                replacement += f":{alias}"
+            if new_alias is not None:
+                logging.info(f"📦 Updating deployment as {new_alias} in {file}")
+                replacement += f":{new_alias}"
             else:
                 logging.info(f"📦 Updating {address} in {file}")
             replacement += "\n"
--- a/src/nile/deployments.py
+++ b/src/nile/deployments.py
@@ -25,16 +25,12 @@ def register(address, abi, network, alias):
         fp.write("\n")
 
 
-def update(address, abi, network, alias):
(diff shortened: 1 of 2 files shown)


ericglau:
Should this function be called `update_abi` for clarity?  I don't think there is any need to update the alias.

andrew-fleming:
> Should this function be named update_abi for clarity? I don't think there is any need to update a deployment's alias.

Yeah, `update_abi` works much better IMO; although, I can see updating an alias as a (separate) useful feature

andrew-fleming (dismissed):
Very nice! Looks good to me

src/nile/deployments.py
```diff
@@ -28,6 +28,57 @@ def register(address, abi, network, alias):
...
+
+    if not os.path.exists(file):
+        raise Exception(f"{file} does not exist")
+
+    with open(file, "r") as fp:
+        lines = fp.readlines()
+
+    found = False
+    for i in range(len(lines)):
+        line = lines[i].strip().split(":")
```
ericnordelo, line 49: If there are multiple aliases, this method removes them except for the first one. I don't know if this behavior is documented, but other methods (like load) assume that multiple aliases can exist (getting all of them in *aliases) from the line. I suggest refactoring this to allow multiple aliases separated with a colon (keep them and recognize them).

ericnordelo (changes requested):
I left a comment for keeping consistency with allowing multiple aliases.

src/nile/deployments.py
ericglau: Changed to handle multiple aliases.

How should multiple aliases be added in the first place though?  (For the tests, I'm assuming they are registered with a string like "alias1:alias2", since the `register` command does not currently allow appending an alias to an existing deployment)

ericnordelo: Issue #220 should add better specifications for this. For this PR, I think maintaining consistency with the existing codebase is enough.

src/nile/deployments.py
```diff
@@ -28,6 +28,58 @@ def register(address, abi, network, alias):
...
+
+            replacement = f"{current_address_hex}:{abi}"
+            if current_aliases is not None:
+                replacement += ":" + ":".join(str(x) for x in current_aliases)
+            replacement += "\n"
+
+            lines[i] = replacement
+
+            found = True
+            break
```
ericnordelo, line 74: We can use the same pattern as in `load` method here I think:
```suggestion
        [address, abi, *aliases] = line.strip().split(":")
        address = normalize_number(address)
        identifiers = [address]
        
        if type(address_or_alias) is not int:
            identifiers = alias
        
        if address_or_alias in identifiers:
            logging.info(f"📦 Updating deployment {address_or_alias} in {file}")
            
            replacement = f"{address}:{abi}"
            if aliases is not None:
                replacement += ":" + ":".join(str(x) for x in aliases)
            replacement += "\n"
            lines[i] = replacement
            found = True
            break
```

ericnordelo (commented):
Looking good! Just added a suggestion to keep code consistency and I think is good to go.

PR #228, opened 2022-10-06
Update comment for more clarity

ericnordelo:


src/nile/cli.py
```diff
@@ -126,7 +126,8 @@ def send(signer, address_or_alias, method, params, network, max_fee=None):
             method, address_or_alias, [x for x in params]
         )
     )
-    # Account.send is part of the public API, so it accepts addresses as string
+    # address_or_alias is not normalized first here, because Account.send
+    # is part of the nile public API, consequently accepting address as hex too
```
andrew-fleming, line 130:
```suggestion
    # address_or_alias is not normalized first here because
    # Account.send is part of Nile's public API and can accept hex addresses
```
Something like this might be a bit easier to follow. What do you think?

ericnordelo: Looks better for sure :)

andrew-fleming (approved):

PR #230, opened 2022-10-06
Add queries to CLI and Account interface

ericnordelo:
Drafted because it lacks test and documentation but is functional already. I will add the rest tomorrow.

It can be used from the NRE, but the methods are returning a string (instead of the estimated gas fee, for example) because the current design of the CLI prints from this output instead of logging properly (output needs to be parsed from NRE).

A refactor to use loggers properly would allow us to return the proper values for NRE use (like tx_hash in Account.send instead of the full string from starknet.cli).


andrew-fleming (dismissed):
Great work, @ericnordelo! I just left a tiny suggestion :) let's also keep in mind that we'll have to add a `version` arg to Contracts's signer methods before releasing Contracts

src/nile/cli.py
```diff
@@ -117,18 +117,26 @@ def setup(signer, network):
 @click.argument("method", nargs=1)
```
martriay, line 117: why the change in the permissions? (`100755` → `100644`)

src/nile/core/account.py
```diff
@@ -79,29 +79,58 @@ def deploy(self):
...
-            max_fee = 0
-        else:
-            max_fee = int(max_fee)
+    def estimate_fee(self, address_or_alias, method, calldata, max_fee, nonce=None):
+        """Estimate fee for a tx going through an Account contract."""
+        return self.execute_query(
+            "estimate_fee", address_or_alias, method, calldata, max_fee, nonce
+        )
+
+    def execute_query(
```
martriay, line 118: what do you think about renaming this `execute` or something similar, and to also accept a `transaction` type that actually executes it, to avoid code duplication with `send`

src/nile/core/call_or_invoke.py
```diff
@@ -10,9 +10,27 @@
...
+    Call or invoke functions of StarkNet smart contracts.
+
+    @param contract: can be an address, an alias, or an Account instance.
+    @param type: can be either call or invoke.
+    @param method: the targeted function.
+    @param params: the targeted function arguments.
+    @param network: goerli, mainnet, or predefined networks file.
+    @param signature: optional signature for invoke transactions.
+    @param max_fee: optional max fee for invoke transactions.
+    @param query_flag: either simulate or estimate_fee
```
martriay, line 32: 👍

tests/commands/test_account.py
```diff
@@ -111,3 +116,69 @@ def test_send_sign_transaction_and_execute():
...
+    # Mock sign_transaction
+    account.signer.sign_transaction = MagicMock(return_value=return_signature)
+
+    account.execute_query(query_type, account.address, "method", [1, 2, 3], max_fee=0)
+
+    account.signer.sign_transaction.assert_called_once_with(
+        calls=[send_args],
+        nonce=0,
+        sender=account.address,
+        max_fee=0,
```
martriay, line 170: do you think we should pass a `max_fee` different than `0` to make sure it's going through, considering `0` is the default?

martriay (approved):
Looks good! Left some comments

src/nile/cli.py
ericnordelo: Not sure at what moment the permission was updated, to be honest. Fixed again to 755.

martriay (approved):
Please resolve conflicts and good to go!

martriay (dismissed):

andrew-fleming (approved):
Great work, @ericnordelo! Looks good to me :)

martriay (approved):

PR #233, opened 2022-10-10
Fix dependency management

### LINKED ISSUE #134: Dependency management is not accurate

andrew-fleming:
This PR proposes to:
- remove `install` call from `init` (if accepted, we can remove `install` altogether)
- remove installations from `init`
- add `starknet-devnet` and `cairo-lang` to setup.cfg's required installations
- adjust supported python versions in setup.cfg

If requiring >=python3.9 is agreed, this PR should also update the CI to account for this.

Resolves #134.


setup.cfg
```diff
@@ -20,12 +20,14 @@ include_package_data = True
...
 
-python_requires = >=3.7
+python_requires = >=3.9,<3.10
 
 install_requires =
     click>=8.0,<9.0
     importlib-metadata>=4.0
     python-dotenv>=0.19.2
+    starknet-devnet>=0.3.1
+    cairo-lang>=0.10.0
```
ericnordelo, line 30: I would add pytest requirement too, because we include a test example in the base project that depends on pytest, and having to install that manually from the user perspective seems not the best UX.

andrew-fleming: > we include a test example in the base project that depends on pytest, and having to install that manually from the user perspective seems not the best UX.

Agreed!^ Though, `cairo-lang` already includes `pytest` and `pytest-asyncio` as deps...I don't know if there's a benefit in explicitly installing it. I suppose the same argument can be made for `cairo-lang` since devnet already has it as a dep

ericnordelo: I think that explicitly installing it helps us to define the version requirement for us, which in many cases could be not too important, but in certain occasions could be. 

**requests** is also a dependency we need to add btw.

andrew-fleming: That's totally reasonable. Thank you 🙏 will update accordingly

README.md
```diff
@@ -33,12 +33,11 @@ nile init
 ...
 ✨  Cairo successfully installed!
 ...
-✅ Dependencies successfully installed
 🗄  Creating project directory tree
 ⛵️ Nile project ready! Try running:
 ```
 
-This command creates the project directory structure and installs `cairo-lang`, `starknet-devnet`, `pytest`, and `pytest-asyncio` for you. The template includes a makefile to build the project (`make build`) and run tests (`make test`).
+This command provides a template with a makefile to build the project (`make build`) and run tests (`make test`).
```
martriay, line 40: actually there's not a makefile anymore

.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       matrix:
         # Add '3.10-dev' once frozendict issue is resolved
         # See https://github.com/starkware-libs/cairo-lang/issues/38
-        python_version: [ '3.8', '3.9' ]
+        python_version: [ '3.9' ]
```
martriay, line 22: what was the need for this?

src/nile/core/init.py
```diff
@@ -1,30 +1,11 @@
...
 
-from nile.core.install import install
-
 
 def init():
     """Kickstart a new Nile project."""
-    # install cairo dependencies
-    subprocess.check_call(
-        [sys.executable, "-m", "pip", "install", "ecdsa", "fastecdsa", "sympy"]
-    )
```
martriay, line 16: are these installed by `cairo-lang` now? it wasn't the case before

src/nile/core/init.py
```diff
@@ -1,30 +1,11 @@
 """Command to kickstart a Nile project."""
 import logging
-import subprocess
-import sys
 from distutils.dir_util import copy_tree
 from pathlib import Path
 
-from nile.core.install import install
```
martriay, line 8: shouldn't we also remove this module?

martriay (dismissed):
good improvements! left some comments

.github/workflows/ci.yml
andrew-fleming: I guess bad assumption on my part? The `cairo-lang` v0.10 release stated they changed from python3.7 to 3.9 ([release](https://github.com/starkware-libs/cairo-lang/releases/tag/v0.10.0)) and their documentation changed their environment setup from using python3.7 to 3.9 (see [quickstart](https://www.cairo-lang.org/docs/quickstart.html)). python3.8 still seems to work (at least with Nile). I suppose we should keep it unless we find it's broken. What do you think?

src/nile/core/init.py
andrew-fleming: Yes, indeed

src/nile/core/init.py
andrew-fleming: Yes! I wanted to make sure the changes were accepted before deleting (and potentially adding it back). Will delete

.github/workflows/ci.yml
martriay: if cairo-lang works, let's maintain it

README.md
martriay: i would mention what the command does tho. it creates the project structure, sample files, etc. maybe we should also create an .env file? that'd be a different issue tho

martriay (approved):

README.md
andrew-fleming: > i would mention what the command does tho

On it

> maybe we should also create an .env file

I think that's a great idea! #249

ericnordelo (approved):
Looks good to me!

PR #236, opened 2022-10-17
Replace subprocess

### LINKED ISSUE #161: Replace subprocess with direct python calls

andrew-fleming:
This PR supersedes #169 and resolves #161.

This PR proposes to include some notable changes that enable the removal of `subprocess`.

### Account
- Initializing the `Account` class requires `Account` to inherit the `AsyncObject` class.
  - Python does not allow asynchronous `__init__`s; therefore, this is a bit of a hacky workaround.
  - A cleaner solution could be to decouple `Account` initialization and deployment.

### starknet_cli
- Direct `starknet_cli` calls accept two parameters: `args` and `command_args`.
  1. `args`: a dictionary wrapped in `SimpleNamespace`.
  2. `command_args`: commands passed in an array as strings (similar to how Nile passed arguments to `subprocess`).
- The new `set_args` function just requires the network and returns the necessary object to pass to `starknet_cli`.
- `command_args` are appropriately created in the function itself.
- The new `capture_stdout` captures the print statements from the `starknet_cli` calls. This is because those function calls do not return the values.

### CLI
- The CLI requires `asyncclick`.

### NRE
- Scripts must now be written with `async`/`await`.
  - `async def run(nre):`
  - `balance = await nre.call(...)`

### Testing
- Tests, of course, must use `async`/`await` when testing async functions.
- For asynchronous mocks, `AsyncMock` should be used.


src/nile/core/deploy.py
```diff
@@ -17,18 +26,27 @@ def deploy(
...
+
+    if arguments:
+        command_args.append("--inputs")
+        command_args.extend(prepare_params(arguments))
+
+    if mainnet_token:
+        command_args.append("--token")
+        command_args.extend(mainnet_token)
+
+    args = set_args(network)
```
martriay, line 46: can't we abstract these out from the core commands even if `run_command` is gone?

src/nile/core/deploy.py
```diff
@@ -17,18 +26,27 @@ def deploy(
...
+
+    output = await capture_stdout(
+        starknet_cli.deploy(args=args, command_args=command_args)
```
martriay, line 49: can't we abstract this with something like:
```python
def call_cli(cmd_name, args, command_args):
    cmd = getattr(starknet_cli, cmd_name)
    return capture_stdout(
        cmd(args=args, command_args=command_args)
    )

**Review by martriay — COMMENTED** [review]
Still looking at it, but good refactor! I think the async/await thing can be slightly annoying but we may be able to get rid of it if we bypass the `starknet_cli` module altogether, which I think we should eventually do to get better error handling and more flexibility in general (e.g. be able to implement a more generic `deploy_account` command with custom account contracts)

File: src/nile/core/deploy.py (hunk 2 above)
Comment (andrew-fleming) [reply src/nile/core/deploy.py:49]: Good call, that should work!

File: src/nile/core/deploy.py (hunk 1 above)
Comment (andrew-fleming) [reply src/nile/core/deploy.py:46]: The short answer is: probably yes. I didn't think it made sense to do so prior to all the recent conflict fixes, but it might now. I'll look into it

File: src/nile/core/deploy.py (hunk 2 above)
Comment (martriay) [reply src/nile/core/deploy.py:49]: does it make sense to abstract `starknet_cli` into its own module exposing the commands?

Comment (andrew-fleming) [reply src/nile/core/deploy.py:49]: The only potential benefit I see right now is removing reliance on common.py (which might be worth doing).

File: src/nile/cli.py [hunk 3]
```diff
@@ -254,32 +261,32 @@ def version():
...
-    return get_nonce_command(normalize_number(contract_address), network)
+    return await get_nonce_command(normalize_number(contract_address), network)
 
 
 cli = load_plugins(cli)
 
 
 if __name__ == "__main__":
-    cli()
+    cli(_anyion_backend="asyncio")
```
martriay, line 292: this results in this being logged every time:
```
Using selector: EpollSelector
```

image


try [disabling the logging](https://stackoverflow.com/a/60503706/1633783)

src/nile/common.py
```diff
@@ -212,3 +227,9 @@ def get_hash(contract_name, overriding_path=None):
     return hex_class_hash(
         compute_class_hash(contract_class=contract_class, hash_func=pedersen_hash)
     )
+
+
+def call_cli(cmd_name, args, command_args):
+    """Make call to starknet_cli and return captured stdout."""
+    cmd = getattr(starknet_cli, cmd_name)
+    return capture_stdout(cmd(args=args, command_args=command_args))
```
martriay, line 235: i think we can abstract all these CLI related functions (gateways, set_args, etc..) into a CLI class that also exposes `invoke` and `call` based on this one. `call_cli` was a proof of concept name but i don't really like it

src/nile/core/account.py
```diff
@@ -26,10 +26,24 @@
 load_dotenv()
 
 
-class Account:
+class AsyncObject(object):
+    """Base class for Account to allow async initialization."""
+
+    async def __new__(cls, *a, **kw):
+        """Return coroutine (not class so sync __init__ is not invoked)."""
```
martriay, line 33: this might not be necessary if we advance with https://github.com/OpenZeppelin/nile/issues/278. maybe worth adding a comment about it

src/nile/core/account.py
```diff
@@ -95,7 +109,7 @@ def declare(
     ):
         """Declare a contract through an Account contract."""
         if nonce is None:
-            nonce = get_nonce(self.address, self.network)
+            nonce = int(await get_nonce(self.address, self.network))
```
martriay, line 112: doesn't `get_nonce` cast to int already?

martriay: wondering if it makes sense to add a `get_nonce` function into this account module, that knows its own address and network

src/nile/core/account.py
```diff
@@ -172,15 +188,19 @@ def send(
             query_flag=query_type,
         )
 
-    def simulate(self, address_or_alias, method, calldata, max_fee=None, nonce=None):
+    async def simulate(
+        self, address_or_alias, method, calldata, max_fee=None, nonce=None
+    ):
         """Simulate a tx going through an Account contract."""
-        return self.send(address_or_alias, method, calldata, max_fee, nonce, "simulate")
+        return await self.send(
```
martriay, line 195: what happens if we don't `await` in here? does it work the same? i.e:
- waiting for `send` if `await simulate()`
- not waiting for anything otherwise

src/nile/core/call_or_invoke.py
```diff
@@ -34,22 +37,37 @@ def call_or_invoke(
     else:
         address, abi = next(deployments.load(contract, network))
 
-    address = hex_address(address)
-    arguments = [
+    args = set_args(network)
+    command_args = set_command_args(
+        inputs=params, signature=signature, max_fee=max_fee, query_flag=query_flag
+    )
+
```
martriay, line 44: probably not now, but i wonder if we could abstract this into the starknet_cli module i'm thinking about

martriay (commented):
left some comments, will continue the review later!

src/nile/core/account.py
andrew-fleming: Good question! `simulate` and `estimate_fee` don't need to be async

src/nile/core/account.py
andrew-fleming: I think it makes sense. It's probably better suited in a separate PR though

src/nile/core/account.py
martriay: they do need to be async. just this line i don't think it needs to be awaited, the user of `simulate()` can `await` when calling it with the same result, no?

tests/commands/test_account.py
```diff
@@ -26,57 +26,68 @@ def tmp_working_dir(monkeypatch, tmp_path):
...
+    expected = [
+        "Account",  # contract
+        [account.signer.public_key],  # arguments
+        NETWORK,  # network
+        f"account-{account.index}",  # alias
+        (
+            f"{test_path}/artifacts",  # overriding-
+            f"{test_path}/artifacts/abis",  # path
+        ),
+    ]
```
martriay, line 69:
```suggestion
    expected = [
        "Account",                   # contract
        [account.signer.public_key], # arguments
        NETWORK,                     # network
        f"account-{account.index}",  # alias
        (                            # overriding-path
            f"{test_path}/artifacts",
            f"{test_path}/artifacts/abis",
        ),
    ]
```

tests/commands/test_debug.py
```diff
@@ -73,14 +74,11 @@ def test__locate_error_lines_with_abis_misformatted_line(mock_path, caplog):
...
 )
-@patch("nile.utils.debug.subprocess.check_output")
-@patch("nile.utils.debug.GATEWAYS", return_value="123")
-def test_debug_feedback_with_message(
-    mock_gateway, mock_output, output, expected, caplog
-):
+async def test_debug_feedback_with_message(output, expected, caplog):
     logging.getLogger().setLevel(logging.INFO)
-    mock_output.return_value = output
+    with patch("nile.utils.debug.call_cli", new=AsyncMock()) as mock_cli_call:
```
martriay, line 79:
```suggestion
    with patch("nile.utils.debug.call_cli") as mock_cli_call:
```
I think `AsyncMock` is the default already?
> If new is omitted, then the target is replaced with an [AsyncMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.AsyncMock) if the patched object is an async function or a [MagicMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.MagicMock) otherwise.

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

tests/commands/test_get_nonce.py
```diff
@@ -1,55 +1,47 @@
...
-    logging.getLogger().setLevel(logging.INFO)
+    # Check return value
+    nonce = await get_nonce(contract_address, network)
+    assert nonce == NONCE
 
-    # check return values
-    nonce = get_nonce("0xffff", "localhost")
-    assert nonce == 5
+    # Check internal
+    mock_without_log.assert_called_once_with(contract_address, network)
```
martriay, line 29: does it make sense to test that `get_nonce` calls an internal function? it seems like testing implementation details. any reason not to have a single test of just behavior as the one below?

tests/conftest.py
```diff
@@ -1 +1,9 @@
 """Test configuration for pytest."""
+import asyncio
+
+import pytest
+
+
+@pytest.fixture(scope="module")
+def event_loop():
+    return asyncio.new_event_loop()
```
martriay, line 9: why is this needed?

tests/commands/test_deploy.py
```diff
@@ -22,78 +22,54 @@ def tmp_working_dir(monkeypatch, tmp_path):
...
-    mock_register, mock_parse, mock_run_cmd, caplog, args, exp_command, exp_abi
-):
+async def test_deploy(mock_register, mock_parse, caplog, args, exp_abi):
     logging.getLogger().setLevel(logging.INFO)
 
-    # check return values
-    res = deploy(*args)
-    assert res == (ADDRESS, exp_abi)
+    with patch("nile.core.deploy.call_cli", new=AsyncMock()) as mock_cli_call:
+        mock_cli_call.return_value = CALL_OUTPUT
```
martriay, line 59: shouldn't we test that `call_cli` (or starknet_cli) is called with the right parameters? why removing the equivalent of `mock_run_cmd.assert_called_once_with`?

tests/test_common.py
```diff
@@ -92,3 +48,63 @@ def test_stringify(args, expected):
...
+    "network, expected",
+    [
+        ("localhost", "http://127.0.0.1:5050/"),
+        ("goerli", "https://alpha4.starknet.io/gateway"),
+        ("mainnet", "https://alpha-mainnet.starknet.io/gateway"),
+    ],
+)
+def test_get_gateway_url(network, expected):
+    if network == "localhost":
+        with patch("nile.common.GATEWAYS") as mock_gateways:
```
martriay, line 63: not sure i understand why `localhost` is different

martriay (commented):
Great work! Left some comments

tests/commands/test_debug.py
martriay: if so, we should probably update the rest of the tests

src/nile/core/account.py
```diff
@@ -195,15 +209,17 @@ def _get_target_address(self, address_or_alias):
...
             max_fee = 0
         else:
             max_fee = int(max_fee)
 
-        return calldata, max_fee, nonce
+        if calldata is not None:
+            calldata = [normalize_number(x) for x in calldata]
+            return max_fee, nonce, calldata
+
+        return max_fee, nonce
```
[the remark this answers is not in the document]
andrew-fleming: I changed the func to play nicely with `account.declare` (which doesn't have calldata). I was thinking of just passing an empty array for the calldata value in `account.declare`:

```
        """Declare a contract through an Account contract."""
        max_fee, nonce, _ = await self._process_arguments(max_fee, nonce, [])
```

...but thought by making calldata optional in `_process_arguments`, it'd be a less awkward function call. Happy to use the example approach^ if preferred.

____

And a giant +1 for using types. That's a `next` or `after` issue IMO

src/nile/core/declare.py
```diff
@@ -23,20 +29,20 @@ def declare(
...
         raise Exception(f"Alias {alias} already exists in {file}")
 
-    arguments = ["--sender", hex_address(sender)]
     max_fee = "0" if max_fee is None else str(max_fee)
 
-    output = run_command(
-        operation="declare",
-        network=network,
+    args = set_args(network)
+    command_args = set_command_args(
```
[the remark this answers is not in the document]
andrew-fleming: Yeahhh I don't disagree. I'll tighten this up

tests/commands/test_debug.py
andrew-fleming: It's strange, most of the tests follow that approach of _not_ using `new=AsyncMock()`. In the few instances `new=` is still used, it's because the returned mock is a `MagicMock`. I have some guesses, but I'm not definitively sure why. 

Here's an example with `test_debug.py`:

```
async def test_debug_feedback_with_message(output, expected, caplog):
    logging.getLogger().setLevel(logging.INFO)
    with patch("nile.utils.debug.call_cli") as mock_cli_call:
        print("mock type: ", type(mock_cli_call))
```

which returns:

```
tests/commands/test_debug.py::test_debug_feedback_with_message...
mock type:  <class 'unittest.mock.MagicMock'>
```

tests/commands/test_get_nonce.py
andrew-fleming: I think it makes sense. It's eliminating any opportunity for the internal func to manipulate the test. It's probably not a big deal in this simple scenario. Happy to change it if preferred

tests/conftest.py
andrew-fleming: We need the event loop to handle the async funcs (like in Contracts's tests)

tests/commands/test_deploy.py
andrew-fleming: Yes, we should. I must have deleted this by mistake during one of the conflict fixes

tests/test_common.py
andrew-fleming: When the network is `localhost`, a dictionary value is returned from the `get` method (`GATEWAYS.get(network)`). The alpha networks don't use a dict to get their url, so we just mock it for `localhost`...we should also add `goerli2` and `integration` in the test

tests/commands/test_get_nonce.py
martriay: yes, i wouldn't test an auxiliary function call. that's an implementation detail.

src/nile/starknet_cli.py
```diff
@@ -0,0 +1,115 @@
...
+    networks = ["localhost", "goerli2", "integration"]
+    if network in networks:
+        return GATEWAYS.get(network)
+    else:
+        network = "alpha-" + network
+        return f"https://{NETWORKS[network]}/feeder_gateway"
+
+
+def _add_args(key, value):
+    return [f"--{key}", *value] if type(value) is list else [f"--{key}", value]
```
martriay, line 115:
```suggestion
    return [f"--{key}", *value] if type(value) is list else [f"--{key}", value]
```
is it enough with `*value` or should we flatten if there's nested lists?

src/nile/starknet_cli.py
```diff
@@ -0,0 +1,115 @@
...
+    stdout = sys.stdout
+    sys.stdout = io.StringIO()
+    await func
+    output = sys.stdout.getvalue()
+    sys.stdout = stdout
+    result = output.rstrip()
+    return result
+
+
+def set_args(network):
```
martriay, line 45:
```suggestion
def set_context(network):
```
this seems more accurate according the description

martriay (commented):
Looking good!

src/nile/core/account.py
```diff
@@ -222,7 +222,9 @@ def _get_target_address(self, address_or_alias):
 
         return target_address
 
-    async def _process_arguments(self, max_fee, nonce, calldata=None):
+    async def _process_arguments(self, calldata, max_fee, nonce):
```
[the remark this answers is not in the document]
andrew-fleming: Ah right. This was the original order (and without the default). I'll switch it back

src/nile/core/call_or_invoke.py
```diff
@@ -41,48 +40,38 @@ async def call_or_invoke(
...
+        logging.info(output)
         if not query_flag and watch_mode:
             transaction_hash = _get_transaction_hash(output)
             return await status(normalize_number(transaction_hash), network, watch_mode)
 
-        if query_flag:
-            logging.info(output)
-
-        return output
+    return output
```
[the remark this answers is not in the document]
andrew-fleming: It's still logging on query, the log command is before the `query_flag` and `watch_mode` conditional. The logic is just consolidated from how it was:

```
    if type != "call" and output:
        logging.info(output)
        if not query_flag and watch_mode:
            transaction_hash = _get_transaction_hash(output)
            return await status(normalize_number(transaction_hash), network, watch_mode)
```

src/nile/core/declare.py
```diff
@@ -33,18 +28,17 @@ async def declare(
...
     max_fee = "0" if max_fee is None else str(max_fee)
 
-    args = set_args(network)
-    command_args = set_command_args(
+    output = await execute_call(
+        "declare",
+        network,
         contract_name=contract_name,
-        signature=signature,
+        signature=prepare_params(signature),
```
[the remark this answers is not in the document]
andrew-fleming: They can and they were. The idea was to support `set_command_args` using a loop for most of the args (meaning less `if` statements and verbosity). That said, they probably should be called in `execute_call`. I'll throw them back in

src/nile/starknet_cli.py
andrew-fleming: Yeah, flattening seems correct

tests/commands/test_get_nonce.py
```diff
@@ -38,10 +37,10 @@ async def test_get_nonce(mock_without_log, contract_address, network, caplog):
...
+    with patch("nile.utils.get_nonce.execute_call", new=AsyncMock()) as mock_cli_call:
         mock_cli_call.return_value = NONCE
-        await get_nonce_without_log(contract_address, "goerli")
+        await get_nonce_without_log(contract_address, NETWORK)
 
-        args = set_args("goerli")
-        command_args = ["--contract_address", "0x4d2"]
-        mock_cli_call.assert_called_once_with("get_nonce", args, command_args)
+        # args = set_args("goerli")
+        command_args = {"contract_address": "0x4d2"}
```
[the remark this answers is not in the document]
andrew-fleming: This test checks that non-0x addresses convert to hex (hence the hardcoded `0x4d2`)

src/nile/core/declare.py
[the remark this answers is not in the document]
ericnordelo: I see `prepare_params` is used for signature inside `execute_call` already but hasn't been removed here right?

src/nile/core/deploy.py
```diff
@@ -19,21 +25,22 @@ def deploy(
...
     register_abi = abi if abi is not None else f"{base_path[1]}/{contract_name}.json"
 
-    output = run_command(
-        operation="deploy",
-        network=network,
+    output = await execute_call(
+        "deploy",
+        network,
         contract_name=contract_name,
+        inputs=prepare_params(arguments),
```
ericnordelo, line 38: Aren't we calling `prepare_params` for inputs inside `execute_call` already?

ericnordelo (commented):
Looking very good Andrew! I left a couple of comments.

src/nile/cli.py
```diff
@@ -2,7 +2,7 @@
 """Nile CLI entry point."""
 import logging
 
-import click
+import asyncclick as click
```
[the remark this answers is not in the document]
andrew-fleming: Great question! I'm honestly not sure. I'll try and test it to find out

tests/test_starknet_cli.py
```diff
@@ -0,0 +1,145 @@
...
+        ("mainnet", "https://alpha-mainnet.starknet.io/feeder_gateway"),
+    ],
+)
+def test_get_feeder_url(network, expected):
+    if network == "localhost":
+        with patch("nile.starknet_cli.GATEWAYS") as mock_gateways:
+            mock_gateways.get = MagicMock(return_value=expected)
+            url = get_feeder_url(network)
+            assert url == expected
+
```
ericnordelo, line 131:
```suggestion
```

ericnordelo (approved):
Looking very good Andrew!

ericnordelo (approved):
LGTM!

src/nile/core/declare.py
```diff
@@ -26,13 +26,13 @@ async def declare(
         file = f"{network}.{DECLARATIONS_FILENAME}"
         raise Exception(f"Alias {alias} already exists in {file}")
 
-    max_fee = "0" if max_fee is None else str(max_fee)
+    max_fee = "0" if max_fee is None else int(max_fee)
```
martriay, line 29:
```suggestion
    max_fee = 0 if max_fee is None else int(max_fee)
```
_types are coming_ 👻

martriay (approved):
🚀 🚀 🚀 🚀 🚀 🚀

martriay (approved):

PR #240, opened 2022-10-20
Improve project setup description

ericnordelo:


README.md
```diff
@@ -501,9 +501,10 @@ OpenZeppelin Nile exists thanks to its contributors. There are many ways you can
 ## Hacking on Nile
 
 Nile uses tox to manage development tasks, you can get a list of
-available task with `tox -av`.
+available task with `tox -av` after installing the package.
 
 - Install a development version of the package with `python -m pip install .`
+- Install tox for development tasks with `python -m pip install tox`
```
andrew-fleming, line 507: It's my understanding that `tox` shouldn't be installed inside a venv; rather, it should be installed just once per system and used to create venvs. Please correct me if I'm mistaken.

If that is the case, I'd suggest we move this as the first bullet point before creating a venv.

andrew-fleming (commented):
Hey, @ericnordelo! CONTRIBUTING looks good! I left a comment and question regarding the README :)

README.md
```diff
@@ -501,9 +501,10 @@ OpenZeppelin Nile exists thanks to its contributors. There are many ways you can
 ## Hacking on Nile
 
 Nile uses tox to manage development tasks, you can get a list of
-available task with `tox -av`.
+available task with `tox -av` after installing the package.
```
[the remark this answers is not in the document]
ericnordelo: I thought the same thing but couldn't find a design that convinced me. Suggestions are welcome!

README.md
ericnordelo: Both ways work. Tox doesn't create a venv for you to play with, but a runtime environment it caches (outside the venv but inside the project, in .tox directory, even if you install it inside the venv). Then it is not exactly used to create venvs, but test environments (or build, etc...) to run something from config (the point is that creating a venv alongside tox still makes sense).

Even while installing tox globally would work nicely, I feel uncomfortable with recommending users to install things outside the venv (it makes things harder to track, in my opinion), but that's a personal feeling, if you think it is better to recommend to install it globally I will update the PR accordingly.

README.md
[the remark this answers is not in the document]
andrew-fleming: ## Hacking on Nile

- Install a development version of the package with `python -m pip install .`
- Install tox for development tasks with `python -m pip install tox`
- Get a list of available tasks with `tox -av`
- Build the package with `tox -e build`
- Format all files with `tox -e format`
- Check files formatting with `tox -e lint`
----
How does something like this look?

README.md
andrew-fleming: > Then it is not exactly used to create venvs, but test environments 

Agreed^

> I feel uncomfortable with recommending users to install things outside the venv (it makes things harder to track, in my opinion)

That's totally fair and I agree that things can become more difficult to track outside the environment. Keeping it as you have it sounds good to me.

andrew-fleming (approved):
Looks good to me!

martriay (approved):

Issue #72, opened 2022-03-07
Support for cairo path cmdline arg for nile compile


milancermak:
I'm running into this need quite often - I would like to pass additional `--cairo_path` directories  to the `nile compile` command. I know I can use `CAIRO_PATH` env var, but there's benefits of having it as a cmdline arg as well.

Would it be of interest to anyone else? If so, I can submit a PR.

martriay:
sounds good to me

martriay:
@comc yes it should

PR #241, opened 2022-10-20
Add cairo_path to compile command (CLI and NRE)

ericnordelo:
Fixes https://github.com/OpenZeppelin/nile/issues/72


tests/commands/test_compile.py
```diff
@@ -144,3 +144,40 @@ def test__compile_auto_account_flag(mock_subprocess, contract_name, flag):
...
+):
+    path = f"path/to/{contract_name}.cairo"
+
+    mock_process = Mock()
+    mock_subprocess.Popen.return_value = mock_process
+
+    # Use named arguments to allow changing the order without modifying tests
+    compile([path], directory=directory, cairo_path=cairo_path)
+
+    returned_subprocess = [
```
martriay, line 169:
```suggestion
    command = [
```
what do you think?

tests/commands/test_compile.py
```diff
@@ -144,3 +144,40 @@ def test__compile_auto_account_flag(mock_subprocess, contract_name, flag):
...
+)
+def test__compile_cairo_path(
+    mock_subprocess, contract_name, directory, cairo_path, expected_internal_cairo_path
+):
+    path = f"path/to/{contract_name}.cairo"
+
+    mock_process = Mock()
+    mock_subprocess.Popen.return_value = mock_process
+
+    # Use named arguments to allow changing the order without modifying tests
```
martriay, line 166: what do you mean by changing the order?

martriay (approved):
Looks good! Left some comments. Interested in @milancermak's thoughts.

tests/commands/test_compile.py
ericnordelo: Order of the arguments in the signature of the function.

martriay: i don't understand why would they change

ericnordelo: I'm not implying they should (or go to) change, but in the case for some reason is done (because it make more sense for any reason to reorganize the arguments), we don't need to look in the rest of the code for the method to update the order. 

It is similar to the pattern for initialization of structs in Solidity (named arguments allowing reorganization of the order of the members of the struct, for potential packaging in this case, but the principle is the same one).

src/nile/core/compile.py
```diff
@@ -52,16 +60,21 @@ def compile(
...
+    directory=None,
+    cairo_path=None,
+    account_contract=False,
+    disable_hint_validation=False,
 ):
     base = os.path.basename(path)
     filename = os.path.splitext(base)[0]
     logging.info(f"🔨 Compiling {path}")
     contracts_directory = directory if directory else CONTRACTS_DIRECTORY
+    cairo_path = cairo_path if cairo_path else contracts_directory
```
milancermak, line 73: Not quite sure I understand the intent behind this. Aren't `--cairo_path` and `--directories` two unrelated concepts?

For example, if I'd want to use nile to compile my `starknet/tokens/*.cairo` files and use libraries I have in `lib/cairo`, I would run `nile compile --directory starknet/tokens --cairo_path lib/cairo`. Would this work as intended?

FWIW, `--cairo_path` is "a list of directories, separated by ":" to resolve import paths." So if you want, you could merge them together instead of overwriting.

ericnordelo: > For example, if I'd want to use nile to compile my `starknet/tokens/*.cairo` files and use libraries I have in `lib/cairo`, I would run `nile compile --directory starknet/tokens --cairo_path lib/cairo`. Would this work as intended?

It will work as intended.

> FWIW, --cairo_path is "a list of directories, separated by ":" to resolve import paths." So if you want, you could merge them together instead of overwriting.

We are not overwriting one with another, we are just setting `contracts_directory` as the default `cairo_path` when the latter is not provided. This allows removing the **contracts** prefix from all the local imports in your project (cairo_path defaults to the project source directory). As you said, those are separated concepts, so it makes sense to keep them separated, as explained next.

> Not quite sure I understand the intent behind this. Aren't --cairo_path and --directory two unrelated concepts?

`directory` (AKA contracts directory) and `cairo path` are two different concepts (that can be unified following a pattern but are not by default).

`directory` is the folder containing the Smart Contracts. This is not for starknet to process but for nile to discover contracts in the project's scope.

`cairo path`, on the other hand, is a set of directories used by the starknet compiler directly (not by nile) to resolve imports in cairo files.

Note that we can avoid passing `cairo path` if we have a project with a contract importing only from external sources (lib modules). Still, we can't avoid setting `directory` if the contracts folder name is distinct from contracts (src, for example). And we can avoid passing `directory` if the contracts folder is named contracts, but we still need to pass cairo path if we are using a structure of imports requiring this.

milancermak: Thanks for the explanation. All good then 👍

andrew-fleming (commented):
The code looks good to me!

I'd consider including an example in the README's `compile` section—maybe even add a brief description of what `cairo_path` does. I'm not sure it's obvious for users. What do you think?

ericnordelo:
> I'd consider including an example in the README's `compile` section—maybe even add a brief description of what `cairo_path` does. I'm not sure it's obvious for users. What do you think?

If the user requires the `cairo_path` flag, this is for advanced usage of the compiler (handling imports), and it probably needs to understand what it is used for, coming from existent starknet tutorials. I think we would be explaining something related to the cairo compiler itself, and I think it could be unnecessary noise in our README, when it is explained in other places (it is already pretty big and need a refactor soon, in my opinion). I think it would be probably helpful to add, after improving our README organization.

andrew-fleming (approved):
> it is already pretty big and need a refactor soon, in my opinion

Yeah, I see what you mean. Agreed.

> I think it would be probably helpful to add, after improving our README organization.

That's totally fair.

tests/commands/test_compile.py
martriay: if it's a design decision i'd add that to contributing and not in here. i vouch for minimal comments when they're really helping understand the code it's being read.

martriay (dismissed):
re: refactoring the readme. I think it follows from #157.

tests/commands/test_compile.py
ericnordelo: Makes sense, I removed the comment.

martriay (approved):

PR #255, opened 2022-10-24
Update universal deployer address

martriay:
Fixes https://github.com/OpenZeppelin/nile/issues/256


andrew-fleming (approved):

src/nile/common.py
```diff
@@ -25,7 +25,7 @@
 QUERY_VERSION = QUERY_VERSION_BASE + TRANSACTION_VERSION
 UNIVERSAL_DEPLOYER_ADDRESS = (
     # subject to change
-    "0x1a8e53128903a412d86f33742d7f907f14ee8db566a14592cced70d52f96222"
+    "0x041a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf"
```
ericnordelo, line 28: This implies that the UDC will have the same address in all the different networks (goerli, goerli2, integrations, mainnet, etc...). If not, should we add this as a config by network?

martriay: It's assumed, yes. The UDC standard defines its deployment through with deployer=0, fixing its address. We could adapt this if it ever changes.

ericnordelo (approved):
LGTM!

PR #257, opened 2022-10-25
make get_hash return hex

martriay:
fixes https://github.com/OpenZeppelin/nile/issues/258

### DIFF
--- a/src/nile/common.py
+++ b/src/nile/common.py
@@ -197,4 +197,6 @@ def get_contract_class(contract_name, overriding_path=None):
 def get_hash(contract_name, overriding_path=None):
     """Return the class_hash for a given contract name."""
     contract_class = get_contract_class(contract_name, overriding_path)
-    return hex(compute_class_hash(contract_class=contract_class, hash_func=pedersen_hash))
+    return hex(
+        compute_class_hash(contract_class=contract_class, hash_func=pedersen_hash)
+    )
--- a/src/nile/common.py
+++ b/src/nile/common.py
@@ -9,7 +9,7 @@ from starkware.crypto.signature.fast_pedersen_hash import pedersen_hash
 from starkware.starknet.core.os.class_hash import compute_class_hash
 from starkware.starknet.services.api.contract_class import ContractClass
 
-from nile.utils import normalize_number, str_to_felt
+from nile.utils import hex_class_hash, normalize_number, str_to_felt
 
 CONTRACTS_DIRECTORY = "contracts"
 BUILD_DIRECTORY = "artifacts"
@@ -197,6 +197,6 @@ def get_contract_class(contract_name, overriding_path=None):
 def get_hash(contract_name, overriding_path=None):
     """Return the class_hash for a given contract name."""
     contract_class = get_contract_class(contract_name, overriding_path)
-    return hex(
+    return hex_class_hash(
         compute_class_hash(contract_class=contract_class, hash_func=pedersen_hash)
     )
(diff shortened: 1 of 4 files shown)


src/nile/common.py
```diff
@@ -197,4 +197,6 @@ def get_contract_class(contract_name, overriding_path=None):
 def get_hash(contract_name, overriding_path=None):
     """Return the class_hash for a given contract name."""
     contract_class = get_contract_class(contract_name, overriding_path)
-    return compute_class_hash(contract_class=contract_class, hash_func=pedersen_hash)
+    return hex(
```
ericnordelo, line 200: I think we must pad the `class_hash`, similar to what we do with addresses. We have an `hex_address` method in utils, we can add another one like `hex_class_hash`. (If we don't pad, short class hashes because of leading 0s, could be as confusing as short addresses because of leading 0s).

ericnordelo (changes requested):

src/nile/common.py
martriay: agree, i considered it but wasn't sure

andrew-fleming: If we pad `get_hash`, we should update `declare`'s class_hash registration for consistency because it's storing the unpadded hex. Also, @ericglau will probably need to adjust one of the hash values in `declare_impl`: https://github.com/OpenZeppelin/openzeppelin-nile-upgrades/blob/main/src/nile_upgrades/declare_impl.py#L18

martriay: i insist in killing the declarations file. it serves no purpose and it's actually annoying when your declare tx fails.

andrew-fleming: That also resolves the inconsistency :)

martriay: @ericnordelo how long class hashes should be? 67?

src/nile/utils/__init__.py
```diff
@@ -127,12 +130,21 @@ def normalize_number(number):
...
     """Return the 64 hexadecimal characters length address."""
     if type(number) == str and number.startswith("0x"):
-        return _pad_hex_to_64(number)
+        return _pad_hex_to(number, ADDRESS_LEN)
+    else:
+        return _pad_hex_to(hex(int(number)), ADDRESS_LEN)
+
+
+def hex_class_hash(number):
+    """Return the 65 hexadecimal characters length class hash."""
```
ericnordelo, line 139: Either change the number here to 63 or change the comment in hex_address's number to 66 to keep consistency (I didn't consider the "0x" prefix as part of the length of the hexadecimal characters when I added the comment).

ericnordelo (commented):
Looking good! I left just a small suggestion regarding a comment.

src/nile/utils/__init__.py
andrew-fleming: +1 to the former

martriay: the current implementation returns a 67 char len string, 0x + 65 chars. The constants at the beginning of the file refer to `n` in the the `0x + n` count. I think the comment is misleading, it should say the full length, including the 0x.

therefore this comment should say 67, and `hex_address`'s 66

src/nile/common.py
martriay: i'm getting 67-length chars on the tests, so it's at least `0x`+65. how do 63chars fit in a `felt`, which can only fit 31bytes?

martriay: i'm getting 69-length chars on the tests, so it's at least `0x`+67. how do 63chars fit in a `felt`, which can only fit 31bytes?

martriay: i wonder why they pad to 64. how are addresses 64 then? that shouldn't fit a felt

ericnordelo (approved):
LGTM!

PR #263, opened 2022-10-25
Support starknet token for mainnet deployments

ericnordelo:


CryptoIndonesia27:
🐋🐋🐋🐋🐋🐋

README.md
```diff
@@ -108,6 +108,8 @@ Creating artifacts/abis/ to store compilation artifacts
 
 > NOTICE: this method doesn't use an account, which will be deprecated very soon as StarkNet makes deployments from accounts mandatory.
 
+> Token for deployments to Alpha MainNet can be set with the `--token` option.
```
andrew-fleming, line 111:
```suggestion
> Token for deployments to Alpha Mainnet can be set with the `--token` option.
```
It looks a little funny with the capital _-N_, no?

README.md
```diff
@@ -173,6 +175,8 @@ Some things to note:
 
 ### `declare`
 
+> Token for declarations to Alpha MainNet can be set with the `--token` option.
```
andrew-fleming, line 178:
```suggestion
> Token for declarations to Alpha Mainnet can be set with the `--token` option.
```

src/nile/cli.py
```diff
@@ -40,6 +40,14 @@ def network_option(f):
     )(f)
 
 
+def mainnet_token_option(f):
+    """Configure TOKEN option for the cli."""
+    return click.option(
+        "--token",
+        help="Used for deploying contracts in Alpha MainNet.",
```
andrew-fleming, line 47:
```suggestion
        help="Used for deploying contracts in Alpha Mainnet.",
```

andrew-fleming (commented):
Looks great, @ericnordelo! I left a few suggestions (basically just one thing). Also, do you think it's worth mentioning the `--token` option in the README?

README.md
ericnordelo: Agree. I used this because I found it like this in Starknet documentation, but I agree it looks funny :).

andrew-fleming (dismissed):
Looks good to me!

tests/test_common.py
```diff
@@ -14,25 +14,60 @@
...
+@pytest.mark.only
+@pytest.mark.parametrize(
+    "operation, signature, max_fee, query_flag, mainnet_token",
+    [
+        ("invoke", None, 0, "simulate", None),
+        ("call", None, None, None, None),
+        ("deploy", None, 5, None, "token"),
+        ("declare", None, 3, "estimate_fee", "token2"),
+    ],
+)
```
martriay, line 26: have you considered doing all permutations? or is it too slow?
```suggestion
@pytest.mark.parametrize('operation', ['invoke', 'call', 'deploy', 'declare'])
@pytest.mark.parametrize('signature', [None])
...
```

martriay (dismissed):
great addition!

tests/test_common.py
ericnordelo: I think that's a great idea! If I add this, it won't be too slow, but I'm not sure what will happen if we adopt this for all the tests. If this makes the test suite too slow in the future, we can refactor this then, I guess. I will add it.

martriay (approved):

PR #265, opened 2022-10-27
Create .env with `nile init` and remove example.env

### LINKED ISSUE #244: remove example.env

### LINKED ISSUE #249: Create .env with `nile init`
It might be helpful to insert a comment illustrating how users should populate the .env i.e. `# PRIVATE_KEY_1 = 1234...`

andrew-fleming:
Resolves #249 and resolves #244.


src/nile/core/init.py
```diff
@@ -14,6 +14,11 @@ def init():
     with open("accounts.json", "w") as file:
         file.write("{}")
 
+    logging.info("🔑  Creating .env file")
+
+    with open(".env", "w") as file:
+        file.write("# ALIAS = PRIVATE_KEY")
```
martriay, line 20: what about adding it to the sample project directory?

martriay (commented):
Good improvement :) . Left just a comment

src/nile/core/init.py
andrew-fleming: Good idea!

src/nile/base_project/.env
```diff
@@ -0,0 +1 @@
+# ALIAS = PRIVATE_KEY
```
martriay, line 1: let's add an actual private key, and a working file

andrew-fleming: Should we keep it commented like: `# PRIVATE_KEY = 0xe3e70682c2094cac629f6fbed82c07cd`?

And what do you mean by a working file?

martriay: i mean that it's ready to be used. why commenting the key?

andrew-fleming: Ohhhh my mistake, I understand. So let's not comment the key :)

martriay: are `0x` keys currently supported?

andrew-fleming: `0x` keys are indeed supported

ericnordelo (approved):
LGTM!

Issue #258, opened 2022-10-25
make `get_hash` return hex


martriay:
in accordance with the internal data type management pattern (#216)

PR #268, opened 2022-10-28
Support Cairo 0.10.1

martriay:
Fixes https://github.com/OpenZeppelin/nile/issues/267. Depends on #236. Borrowed some code from https://github.com/OpenZeppelin/nile/pull/169.

This PR:
- adds support for counterfactual account deployments, sending a `DeployAccount` transaction directly to starknet's Gateway module
- makes counterfactual deployments the default for `nile setup`
- updates the Account artifact to support `__validate_deploy__`
- adds `nile get-balance` to CLI and NRE, for which adds the ERC20 abi to interact with the Ether contract
- adds `nile conterfactual-address` to precompute account's addresses


src/nile/cli.py
```diff
@@ -90,9 +90,11 @@ def run(path, network):
 @click.option("--alias")
 @click.option("--abi")
 @mainnet_token_option
-def deploy(artifact, arguments, network, alias, abi=None, token=None):
+async def deploy(artifact, arguments, network, alias, abi=None, token=None):
```
ericnordelo, line 93:
```suggestion
async def deploy(artifact, arguments, network, alias, abi, token):
```

We can remove default None values in click commands arguments because this is the default behavior.

src/nile/cli.py
```diff
@@ -2,7 +2,7 @@
 """Nile CLI entry point."""
 import logging
 
-import click
+import asyncclick as click
```
ericnordelo, line 5: Using asyncclick will add the requirement for plugins as well?

ericnordelo (commented):
Looking good! I left a couple of comments.

andrew-fleming (commented):
Looking good! I left a question, and I'll re-review once conflicts are resolved and some of the noise is removed from the blocking PR

src/nile/cli.py
martriay: not sure @andrew-fleming

ericnordelo (dismissed):
Looking good!

ericnordelo (approved):

martriay:
This is lacking docs. On it.

tests/commands/test_account.py
```diff
@@ -263,23 +286,29 @@ async def test_simulate(mock_deploy):
 
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize(
-    "query_type, watch_mode", [("estimate_fee", "track"), ("simulate", "debug")]
-)
-@patch("nile.core.account.deploy", return_value=(MOCK_ADDRESS, MOCK_INDEX))
+@pytest.mark.parametrize("query_type", ["estimate_fee", "simulate"])
+@pytest.mark.parametrize("watch_mode", ["track", "debug"])
```
andrew-fleming, line 290: 👍

andrew-fleming (commented):
Looking good, @martriay! I left a few comments and suggestions :)

andrew-fleming:
With the addition of the `salt` param in `Account`, we should pass `predeployed_info` as a keyword argument in `_check_and_return_account` in get_accounts.py:

```
async def _check_and_return_account(signer, pubkey, network, predeployed_info=None):
    account = await Account(signer, network, predeployed_info=predeployed_info)
    ...
```

https://github.com/OpenZeppelin/nile/blob/support-cairo-0.10.1/src/nile/utils/get_accounts.py#L88

README.md
```diff
@@ -161,11 +161,28 @@ nile setup <private_key_alias>
...
 A few things to note here:
 
+- This is a counterfactual deployment, meaning the deployed account pays for its own deployment. You can use [`nile counterfactual-address`](#counterfactual-address) to predict the account address and send the necessary funds beforehand.
 - `nile setup <private_key_alias>` looks for an environment variable with the name of the private key alias.
-- This creates or updates `localhost.accounts.json` file storing all data related to accounts management.
+- This creates or updates `localhost.accounts.json` file storing all data related to account management.
 - The creates or updates `localhost.deployments.txt` file storing all data related to deployments.
 - `--track` and `--debug` flags can be used to watch the status of the account deployment transaction. See `status` below for a complete description.
 
+### `counterfactual-address`
```
ericnordelo, line 170: Is this specifically for UDC? If that's the case, I think we should mention it, and if not, how can you predict the address if the deployer can define the generation mechanism itself? (by changing the salt like UDC).

ericnordelo: Also is the "counterfactual" part of the command name coming from usual terminology, wouldn't be easier to understand something like "compute-address" (like in [solidity contracts Create2](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/99589794db43c8b285f5b3464d2e0864caab8199/contracts/utils/Create2.sol#L48))

README.md
```diff
@@ -431,6 +448,14 @@ async def run(nre):
     await bob.send(...)
 ```
 
+### `get-balance`
+
+Retrieves the Ether balance for a given contract address.
+
+```sh
+nile get-balance <contract_address> [--network]
```
ericnordelo, line 456:
```suggestion
nile get-balance <contract_address> [--network NETWORK]
```

For consistency.

src/nile/cli.py
```diff
@@ -145,11 +148,25 @@ async def declare(
...
+
+
+@cli.command()
+@click.argument("signer", nargs=1)
+@click.option("--salt", nargs=1, default=None)
+async def counterfactual_address(signer, salt):
+    """Precompute the address of an Account contract."""
+    _signer = Signer(normalize_number(os.environ[signer]))
+    address = hex_address(
+        get_counterfactual_address(salt, calldata=[_signer.public_key])
```
ericnordelo, line 167: To follow up the above comment, maybe `compute_address` instead of `get_counterfactual_address`?

src/nile/common.py
```diff
@@ -21,6 +22,8 @@
 TRANSACTION_VERSION = 1
 QUERY_VERSION_BASE = 2**128
 QUERY_VERSION = QUERY_VERSION_BASE + TRANSACTION_VERSION
+ETH_ABI = f"{pt}/artifacts/abis/ERC20.json"
+ETH_ADDRESS = "0x49D36570D4E46F48E99674BD3FCC84644DDD6B96F7C741B1562B82F9E004DC7"
```
ericnordelo, line 26:
```suggestion
WRAPPED_ETH_ADDRESS = "0x49D36570D4E46F48E99674BD3FCC84644DDD6B96F7C741B1562B82F9E004DC7"
```

I was confused for the purpose of this ETH_ADDRESS at first glance.

src/nile/common.py
```diff
@@ -134,6 +137,13 @@ def get_hash(contract_name, overriding_path=None):
     )
 
 
+def get_account_hash(contract="Account"):
+    """Return the class_hash of an Account contract."""
+    pt = os.path.dirname(os.path.realpath(__file__)).replace("/core", "")
+    overriding_path = (f"{pt}/artifacts", f"{pt}/artifacts/abis")
+    return int(get_hash(contract, overriding_path=overriding_path), 16)
```
ericnordelo, line 144:
```suggestion
    return int(get_class_hash(contract, overriding_path=overriding_path), 16)
```
There are a lot of hashes in our scope, I think renaming `get_hash` to `get_class_hash` makes reading the code easier.

ericnordelo: Also, `get_hash` should return an int following the hex/int pattern (I changed these two things in the PR for deploying through accounts).

ericnordelo: I would rename `get_account_hash` to `get_account_class_hash` accordingly.

tox.ini
```diff
@@ -16,8 +16,8 @@ extras =
 deps =
     anyio>=3.6.1
     asyncclick>=8.1.3.4
-    cairo-lang==0.10.0
-    starknet-devnet==0.3.1
+    cairo-lang==0.10.1
```
ericnordelo, line 19: If we have to do nothing to support 0.10.2, why not support 0.10.2?

ericnordelo (commented):
Looking good! Left a few comments.

README.md
martriay: as stated a few lines above, this function is to know the address of a counterfactual deployment in advance, to let the user send the funds. it's a specific case of the more general address computation function (defaults to deployed from 0, picks up the signer public key, etc) so i believe leaving the counterfactual word there helps understand the purpose, instead of making it broader ("compute-address").

src/nile/common.py
martriay: what caused the confusion?

I favor ETH over WRAPPED_ETH because:
- you cannot wrap nor unwrap it
- unlike in Ethereum, there's only one of them here in StarkNet
- and the convention seems to be call it Ether:
image

src/nile/common.py
martriay: the [last understanding](https://github.com/OpenZeppelin/nile/issues/258) was that `get_hash` was considered part of the external API.

In any case I think this pattern is starting to show some flaws: it happens a lot that "external" functions are used internally, and it's also very hard to know in advance whether a function is/will be external or not. It's not something you can derive from a functions name, module, etc. It's very hard to be mindful of the pattern. I wonder if we can do something to fix it.

tox.ini
martriay: this is just for internal tests, the package will support 0.10.2 immediately even with this. it's the same, no strong opinions on bumping it.

src/nile/common.py
ericnordelo: In this case, it doesn't matter if it is internal or external API. The pattern is to return always int unless the function name explicitly declares the intention of returning a hex.

src/nile/common.py
ericnordelo: WRAPPED was perhaps not the best suggestion (maybe ETHER_ADDRESS? ETH_TOKEN_ADDRESS?). 

The confusion was with the purpose of ETH_ADDRESS (it wasn't clear to me that it was for an ERC20 token, I thought it could have been a bridge or something different until I saw the ABI, keeping in mind that ETH is used for Ethereum a lot, like in `EthAccount.cairo` in cairo-contracts, where is not representing the Ether Token itself). 

But was a small suggestion, ETH_ADDRESS is good enough.

tox.ini
ericnordelo: That's true for tox.ini, but to support 0.10.2 immediately, we need to bump the version in `setup.cfg` to 0.10.2 instead of 0.10.1, right? If not installing nile would bring 0.10.1, not 0.10.2. We are using low-level functions from cairo-lang now though, no sure if something will break from its internal API.

src/nile/common.py
martriay: ah i get it now. yes it makes sense.

src/nile/signer.py
```diff
@@ -105,15 +136,17 @@ def get_declare_hash(sender, contract_class, max_fee, nonce, chain_id):
...
-    return calculate_transaction_hash_common(
-        tx_hash_prefix=prefix,
-        version=version,
-        contract_address=account,
-        entry_point_selector=0,
-        calldata=calldata,
+def get_deploy_account_hash(
+    contract_address, class_hash, calldata, salt, max_fee, nonce, chain_id
+):
+    """Compute the hash of a declare transaction."""
```
andrew-fleming, line 142:
```suggestion
    """Compute the hash of an account deployment transaction."""
```

src/nile/common.py
andrew-fleming: Seconded on the initial uncertainty with `ETH_ADDRESS`. I think `ETH_ABI` right next to it implies what it is, but maybe a comment would be helpful. Just a thought

andrew-fleming (commented):
Excellent improvements! This PR is shaping up really well. I left a few minor comments and questions

tox.ini
andrew-fleming: I agree with Eric's comment in keeping the setup.cfg and tox.ini consistent. That said, I just tried installing cairo-lang v0.10.2 and I'm on 15 minutes so far with pip checking version compatibility. Here's a small glimpse:

```
Using cached certifi-2022.5.18.1-py3-none-any.whl (155 kB)
  Using cached certifi-2021.10.8-py2.py3-none-any.whl (149 kB)
INFO: pip is looking at multiple versions of certifi to determine which version is compatible with other requirements. This could take a while.
  Using cached certifi-2021.5.30-py2.py3-none-any.whl (145 kB)
  Using cached certifi-2020.12.5-py2.py3-none-any.whl (147 kB)
  Using cached certifi-2020.11.8-py2.py3-none-any.whl (155 kB)
  Using cached certifi-2020.6.20-py2.py3-none-any.whl (156 kB)
  Using cached certifi-2020.4.5.2-py2.py3-none-any.whl (157 kB)
INFO: This is taking longer than usual. You might need to provide the dependency resolver with stricter constraints to reduce runtime. If you want to abort this run, you can press Ctrl + C to do so. To improve how pip performs, tell us what happened here: https://pip.pypa.io/surveys/backtracking
  Using cached certifi-2020.4.5.1-py2.py3-none-any.whl (157 kB)
  Using cached certifi-2020.4.5-py2.py3-none-any.whl (156 kB)
  Using cached certifi-2019.11.28-py2.py3-none-any.whl (156 kB)
  Using cached certifi-2019.9.11-py2.py3-none-any.whl (154 kB)
  Using cached certifi-2019.6.16-py2.py3-none-any.whl (157 kB)
  Using cached certifi-2019.3.9-py2.py3-none-any.whl (158 kB)
  Using cached certifi-2018.11.29-py2.py3-none-any.whl (154 kB)
  Using cached certifi-2018.10.15-py2.py3-none-any.whl (146 kB)
  Using cached certifi-2018.8.24-py2.py3-none-any.whl (147 kB)
  Using cached certifi-2018.8.13-py2.py3-none-any.whl (146 kB)
  Using cached certifi-2018.4.16-py2.py3-none-any.whl (150 kB)
  Using cached certifi-2018.1.18-py2.py3-none-any.whl (151 kB)
  Using cached certifi-2017.11.5-py2.py3-none-any.whl (330 kB)
  Using cached certifi-2017.7.27.1-py2.py3-none-any.whl (349 kB)
  Using cached certifi-2017.7.27-py2.py3-none-any.whl (349 kB)
  Using cached certifi-2017.4.17-py2.py3-none-any.whl (375 kB)
INFO: pip is looking at multiple versions of requests to determine which version is compatible with other requirements. This could take a while.
Collecting requests>=2.28.1
```

I'll update when this process finishes, but I suggest we keep v0.10.1 for now

andrew-fleming (dismissed):
Great work! :)

andrew-fleming (approved):

src/nile/cli.py
```diff
@@ -183,11 +186,25 @@ async def declare(
 
 @cli.command()
 @click.argument("signer", nargs=1)
+@click.option("--salt", nargs=1)
+@click.option("--max_fee", nargs=1)
 @network_option
 @watch_option
-async def setup(signer, network, watch_mode):
+async def setup(signer, network, salt, max_fee, watch_mode):
```
ericnordelo, line 193: Hi Marto, I was trying to test the setup of an account, and it looks like I can't `setup` an account because the `class_hash` is not declared, and I can't declare it because I have no account deployed to declare through. Am I missing something?

src/nile/core/account.py
```diff
@@ -79,32 +91,55 @@ async def __init__(self, signer, network, predeployed_info=None, watch_mode=None
             self.address = signer_data["address"]
             self.index = signer_data["index"]
         else:
-            address, index = await self.deploy(watch_mode=watch_mode)
-            self.address = address
-            self.index = index
+            output = await self.deploy(
+                salt=salt, max_fee=max_fee, watch_mode=watch_mode
+            )
+            if output is not None:
```
ericnordelo, line 97: If this `output` is None, `self.address` is never assigned, and the assertion below reverts (is not an invariant because of this). 

This is an issue when a deploy account transaction reverts because instead of showing just the error from the status command, the execution raises an exception from the assertion because self.address is not defined.

I suggest changing the assertion to this to keep the message clear:

```
if hasattr(self, 'address'):
```

To visualize the issue just run `nile setup account` after running a clean node with `nile node`.

ericnordelo (commented):
We are almost there! Left just a suggestion and a question.

ericnordelo (approved):
Looking good!

PR #270, opened 2022-10-28
Refactor tests and change network in tests to `localhost`

### LINKED ISSUE #269: Too many requests (HTTP error 429) in tests
Some of the test modules use `goerli` as the network and make calls to the testnet API. Some tests now fail because of Error 429 (too many requests). This is especially a problem with `test_account.py`. To rectify this, we should consider using a mock account instance.

andrew-fleming:
Resolves #269.


martriay (approved):
Good solution

ericnordelo (approved):
LGTM!

PR #272, opened 2022-10-28
Add support for goerli2 and integration testnets

### LINKED ISSUE #271: Support integration and goerli2 testnets

martriay:
Fixes #271


src/nile/common.py
```diff
@@ -38,7 +38,13 @@ def get_gateway():
...
     except FileNotFoundError:
         with open(NODE_FILENAME, "w") as f:
-            f.write('{"localhost": "http://127.0.0.1:5050/"}')
+            f.write(
+                """{
+                "localhost": "http://127.0.0.1:5050/",
+                "goerli2": "https://alpha4-2.starknet.io",
+                "integration": "https://external.integration.starknet.io"
+            }"""
+            )
```
andrew-fleming, line 47:
```suggestion
            f.write(
"""{
    "localhost": "http://127.0.0.1:5050/",
    "goerli2": "https://alpha4-2.starknet.io",
    "integration": "https://external.integration.starknet.io"
}"""
            )
```

The json looks a little funky. The suggestion will look like this:
```
{
    "localhost": "http://127.0.0.1:5050/",
    "goerli2": "https://alpha4-2.starknet.io",
    "integration": "https://external.integration.starknet.io"
}
```

tests/test_cli.py
```diff
@@ -181,6 +181,8 @@ def test_node_runs_gateway(opts, expected):
     "args",
     [
         ([MOCK_HASH, "--network", "goerli"]),
+        ([MOCK_HASH, "--network", "goerli2"]),
+        ([MOCK_HASH, "--network", "integration"]),
```
andrew-fleming, line 185: Since these aren't yet supported by the `STARKNET_ENVIRONMENT`, the subprocess call includes the explicit `--feeder_gateway_url`. These tests should pass if you add the url to `expected` for the goerli2 and integration network tests.

andrew-fleming (approved):
Looks great, @martriay! I left a couple small comments :)

src/nile/cli.py
```diff
@@ -43,7 +43,7 @@ def network_option(f):
 def _validate_network(_ctx, _param, value):
     """Normalize network values."""
     # normalize goerli
-    if "goerli" in value or "testnet" in value:
+    if "goerli" == value or "testnet" in value:
```
ericnordelo, line 46:
```suggestion
    if "testnet" == value:
```
The check for equality to "goerli" is unnecessary because "goerli" is part of the NETWORKS. I would also remove the "in" check in favor of equality, because recognizing for example "testnet-2" as goerli, instead of goerli2 seems a little confusing.

src/nile/cli.py
```diff
@@ -43,7 +43,7 @@ def network_option(f):
...
         return "goerli"
     # normalize localhost
     if "localhost" in value or "127.0.0.1" in value:
```
ericnordelo, line 49:
```suggestion
    if "127.0.0.1" == value:
```

I think accepting something like "localhost15", or "3127.0.0.12" as localhost is not the best approach, I would remove the "in" check.

src/nile/cli.py
ericnordelo: In most cases, I think the user will be passing values matching NETWORKS elements, so I would move this on top of this check, to avoid two ifs that won't enter in most cases. 

```
# check if value is accepted
if value in NETWORKS:
    return value
```

src/nile/common.py
ericnordelo, line 47:
```suggestion
            networks = {
                "localhost": "http://127.0.0.1:5050/",
                "goerli2": "https://alpha4-2.starknet.io",
                "integration": "https://external.integration.starknet.io"
            }
            f.write(json.dumps(networks, indent=2))
```

I think this makes the code look cleaner.

ericnordelo (commented):
Looking good! Left a couple of suggestions.

src/nile/common.py
ericnordelo:
```suggestion
            networks = {
                "localhost": "http://127.0.0.1:5050/",
                "goerli2": "https://alpha4-2.starknet.io",
                "integration": "https://external.integration.starknet.io"
            }
            f.write(json.dumps(networks, indent=2))
```

I think this makes the code look cleaner.

martriay: i'll go with eric's

andrew-fleming (approved):
Looks ready to me!

ericnordelo (approved):
Looks good to go!

PR #275, opened 2022-11-02
support mainnet chain_id

### LINKED ISSUE #234: Add `chain_id` param for signer methods
Currently, nile.signer's `sign_transaction` (and soon-to-be `sign_declare`) only supports StarkNet's testnet chain id. The signer methods should include a parameter to adjust the chain id based on the network.

martriay:
Fixes #234.
Nile is no longer a test util.


andrew-fleming (approved):
Very nice :)

ericnordelo (approved):
Looking good!

Issue #212, opened 2022-09-23
Improve account management


martriay:
Today there's an implied 1:1 relationship between accounts, singers, and private keys. I think we should design better around an alias system that also supports multi-signer, and signers decoupled from private keys to allow rotation.

Decoupling accounts from signers can also help us reuse signers across networks, and set defaults like `max_fee` for an improved UX, besides fixing awkwardness like passing a private key alias to `send` or `declare`.

Issue #73, opened 2022-03-08
Support `nilerc`


fracek:
This is related to #72 and something I personally wish was included. I think it would be helpful to have a configuration file that nile can use to lookup information such as:

 * cairo path
 * contracts directory

For example, the OZ `cairo-contracts` repository contains the cairo contracts in the `openzeppelin` directory and so `nile compile` won't compile them.

I'm also using the OZ contracts in my project and at the moment I can't compile with nile because I can't add the git submodule to the cairo path.

milancermak:
Absolutely spot on. Something I wish for as well.

I could also be a better solution to #75 (which, personally, I don't think fits the scope of nile).

What format should it have? TOML? JSON?

What attributes? So far, we got 1) cairo path, 2) contract directory (directories?), 3) method (starknet vs. cairo).

fracek:
I think TOML is a good format, but a lot of Python projects use Python files for configuration and load them at runtime.

sambarnes:
A lot of the modern python ecosystem seems to be moving towards using the `pyproject.toml` for configuration like this. Coming from other common tools like [poetry](https://python-poetry.org/docs/pyproject/) or [black](https://github.com/psf/black#configuration), that's where I'd likely expect to find options like this (under the assumption we'll continue to have a strong tie to the python ecosystem for as long as nile is relevant)

fracek:
Another option is to have a `nile.toml` file, if we don't want to be so integrated with the Python ecosystem.

Let's also make sure the configuration file allows users to define options for plugins (once implemented, see #53).

martriay:
Is there any advantage of using `nile.toml` above `pyproject.toml` though?

fracek:
With `pyproject.toml` we tie nile to the Python ecosystem. I think that's fine.

milancermak:
I don't think having it in `pyproject.toml` is a good idea. It's a file intended for python packages, it doesn't have anything to do with Nile. I've noticed a lot of people coming to the StarkNet ecosystem (and using nile) have JS dev background and don't have experience with python. Using a python specific config file would confuse them even further (that's why I'd also rather not use a .py file either).

martriay:
Being a JS developer forced into python because of Cairo I can confirm I don't know how to use these files. That being said, I'm not sure how feasible it really is to move away from it, since Nile's scripting API is already python. I don't have a strong opinion, I guess we can just have a `nile.toml`?

PR #276, opened 2022-11-02
Deploy through Accounts

### LINKED ISSUE #259: make `nile deploy` go through an account

ericnordelo:
Fixes #259, #277


README.md
```diff
@@ -113,25 +113,28 @@ Creating artifacts/abis/ to store compilation artifacts
 
 ### `deploy`
 
-> NOTICE: this method doesn't use an account, which will be deprecated very soon as StarkNet makes deployments from accounts mandatory.
+> NOTICE: Calling this method without the `--account` parameter doesn't use an account, which will be deprecated very soon as StarkNet makes deployments from accounts mandatory.
```
andrew-fleming, line 116: The spirit of this is perfect and necessary, but I'd rephrase it a bit. Maybe something like:
```suggestion
> NOTICE: Calling this method will require an account (using the `--account` param) very soon as StarkNet will make deployments from accounts mandatory. 
```

src/nile/cli.py
```diff
@@ -84,15 +84,47 @@ def run(path, network):
...
 @click.option("--abi")
+@click.option("--udc")
 @mainnet_token_option
-def deploy(artifact, arguments, network, alias, abi=None, token=None):
+@network_option
+def deploy(
+    contract_name,
+    salt,
+    params,
+    max_fee,
```
andrew-fleming, line 102: Shouldn't `max_fee` default to `None` here? I don't think it's passed to `account.deploy_contract`

andrew-fleming (commented):
Looking really good, @ericnordelo! I left a few comments.

I do, however, think we should add some tests for `deploy_contract` from deploy.py and account.py. We can probably reuse most of the existing `deploy` tests with some adjustments.

src/nile/core/account.py
```diff
@@ -124,14 +118,32 @@ def declare(
...
+
+        deploy_contract(
+            self,
+            contract_name,
+            salt,
+            unique,
+            calldata,
+            self.network,
+            alias,
+            deployer_address or UNIVERSAL_DEPLOYER_ADDRESS,
```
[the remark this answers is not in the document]
ericnordelo: Good catch!

src/nile/cli.py
ericnordelo: Correct me if I'm wrong, I think that Click behavior is passing the value as None without us needing to set the default in the method (adding unnecessary verbosity IMO). I tested it and this seems to be the behavior (couldn't find the entry in the docs though).

andrew-fleming: It seems you are correct! So yeah, let's remove the unnecessary `None`s

src/nile/core/deploy.py
```diff
@@ -39,3 +54,52 @@ def deploy(
...
+    base_path = (
+        overriding_path if overriding_path else (BUILD_DIRECTORY, ABIS_DIRECTORY)
+    )
+    register_abi = abi if abi is not None else f"{base_path[1]}/{contract_name}.json"
+    deployer_for_address_generation = 0
+
+    if unique:
+        # Match UDC address generation
+        salt = compute_hash_chain(data=[account.address, salt])
+        deployer_for_address_generation = deployer_address
```
andrew-fleming, line 83: What do you think about including a random salt generator for deployments without `--salt` like StarkWare does? => https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/starknet/cli/starknet_cli.py#L846-L847

Otherwise, trying to deploy multiple instances of the same contract will result in:

`raise StarkException(code=code, message=message)
starkware.starkware_utils.error_handling.StarkException: (500, {'code': <StarknetErrorCode.CONTRACT_ADDRESS_UNAVAILABLE: 1>, 'message': 'Requested contract address 1999693730423262048654525999365022023777068478432229193013554800567845966099 is unavailable for deployment.'})`

Users _can_ add `salt` to avoid this, but automating salt seems preferable

andrew-fleming: If we agree^, I definitely suggest adding some salt tests

andrew-fleming (commented):
Looking good! I left some comments :)

Some other notes:
- I think we need to remove line `#85` in deployments: https://github.com/ericnordelo/nile/blob/feat/deploy-through-account-%23259/src/nile/deployments.py#L84-L85
- It might be a good idea to document the requirement of declaring contracts prior to deploying them. At this early stage, I'm not sure it's obvious to do so.

src/nile/core/deploy.py
```diff
@@ -39,3 +54,52 @@ def deploy(
...
+
+    class_hash = get_class_hash(contract_name=contract_name)
+
+    address = calculate_contract_address_from_hash(
+        salt, class_hash, calldata, deployer_for_address_generation
+    )
```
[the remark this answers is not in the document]
ericnordelo: Indeed. `get_class_hash` should return an int to comply with the hex/int pattern we've just documented. @martriay I decided (open to discussion of course) to require returning values as int (not hex), because if not, we would be basically parsing in many parts of the code, because internal API requires to receive ints. Note this doesn't affect output files, or logging, just function return values.

src/nile/core/deploy.py
ericnordelo: I think this salt should be part of the protocol (like nonce in Ethereum deployments), but I am not sure why it is left to be passed from the end users (am I missing an important reason?). With this said, agreeing on a format for the automatic generation of this salt, I think, is content enough for another issue/PR. Shouldn't be any problem with releasing this version, right? I don't see a reason to delay it to another milestone.

ericnordelo: Fixed.

src/nile/core/deploy.py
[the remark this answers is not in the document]
ericnordelo: Fixed.

src/nile/core/deploy.py
andrew-fleming: > I think this salt should be part of the protocol

I agree in general haha but since we're deploying contracts through an `account.send` tx, the burden seems to fall on Nile. When Nile interacts with the starknet_cli directly (and when Nile most likely bypasses the cli in the future), I imagine defining a salt generator will not be necessary

> agreeing on a format for the automatic generation of this salt, I think, is content enough for another issue/PR

That sounds reasonable to me

> I don't see a reason to delay it to another milestone.

I don't disagree

README.md
```diff
@@ -113,25 +113,28 @@ Creating artifacts/abis/ to store compilation artifacts
...
 ### `deploy`
 
-> NOTICE: this method doesn't use an account, which will be deprecated very soon as StarkNet makes deployments from accounts mandatory.
+> NOTICE: Calling this method will require an account very soon (using the `--account` param) as StarkNet will make deployments from accounts mandatory.
 
 > Token for deployments to Alpha Mainnet can be set with the `--token` option.
 
 ```sh
-nile deploy contract --alias my_contract
+nile deploy contract --account setup_account_alias --alias my_contract
```
martriay, line 121: although i like the `--account` flag much better, for consistency with other commands we should use pk aliases
```sh
nile deploy PK_ALIAS contract ...
```
#212 partially addresses it. having an `--account` flag is also the way because it would maintain the API even after having default accounts with https://github.com/OpenZeppelin/nile/issues/73

README.md
```diff
@@ -113,25 +113,28 @@ Creating artifacts/abis/ to store compilation artifacts
...
-1. `nile deploy <contract_name>` looks for an artifact with the same name
-2. This created a `localhost.deployments.txt` file storing all data related to my deployment
-3. The `--alias` parameter lets me create a unique identifier for future interactions, if no alias is set then the contract's address can be used as identifier
-4. By default Nile works on local, but you can use the `--network` parameter to interact with `mainnet`, `goerli`, `goerli2`, `integration`, and the default `localhost`.
-5. By default, the ABI corresponding to the contract will be registered with the deployment. To register a different ABI file, use the `--abi` parameter.
+1. `nile deploy <contract_name>` looks for an artifact with the same name.
+2. This creates a `localhost.deployments.txt` file storing all data related to my deployment.
+3. The `--account` parameter deploys using the provided account and the UniversalDeployer contract.
+4. The `--alias` parameter lets you create a unique identifier for future interactions, if no alias is set then the contract's address can be used as identifier.
+5. The `--udc_address` parameter lets you specify the UniversalDeployer address if needed.
```
martriay, line 135:
```suggestion
5. The `--udc_address` parameter lets you specify the UniversalDeployer address if needed.
```
i think it should be a more generic "deployer address" name in here. not all deployer contracts will be UDC compliant.

README.md
martriay: i understand this makes much more difficult to identify therefore support both deployments (with/without account), unclear what's best here

src/nile/cli.py
```diff
@@ -84,15 +84,47 @@ def run(path, network):
...
+        account = Account(account, network)
+        account.deploy_contract(
+            contract_name,
+            salt,
+            unique,
+            params,
+            alias,
+            deployer_address=udc_address,
+            max_fee=max_fee,
+            abi=abi,
```
martriay, line 122: no `mainnet_token`?

src/nile/core/account.py
```diff
@@ -64,6 +65,9 @@ def __init__(self, signer, network, predeployed_info=None):
             self.address = address
             self.index = index
 
+        # Internal address must always be of type int
+        assert type(self.address) == int
```
martriay, line 69: is this considered internal? i expect `nre` users to access this attribute often

src/nile/core/deploy.py
```diff
@@ -15,8 +26,12 @@ def deploy(
     abi=None,
     mainnet_token=None,
 ):
-    """Deploy StarkNet smart contracts."""
-    logging.info(f"🚀 Deploying {contract_name}")
+    """Deploy StarkNet smart contracts (DEPRECATED)."""
+    logging.info(
+        f"🚀 Deploying {contract_name} without Account. "
+        + "This method is deprecated and will be removed soon. "
+        + "Try using the --account command option."
```
martriay, line 33:
```suggestion
        + "Use the --account option."
```
although this is not accurate for nre

src/nile/core/account.py
```diff
@@ -16,6 +16,7 @@
 from nile.core.call_or_invoke import call_or_invoke
 from nile.core.declare import declare
 from nile.core.deploy import deploy
+from nile.core.deploy import deploy_contract as deploy_with_udc
```
martriay, line 19:
```suggestion
from nile.core.deploy import deploy_contract as deploy_with_deployer
```
for consistency with the above suggestion

tests/commands/test_account.py
```diff
@@ -120,6 +122,43 @@ def test_declare(mock_declare, mock_get_class, mock_deploy):
...
+    max_fee = 1
+
+    account.deploy_contract(
+        contract_name, salt, unique, calldata, alias, max_fee, deployer_address
+    )
+
+    if deployer_address is None:
+        deployer_address = normalize_number(UNIVERSAL_DEPLOYER_ADDRESS)
+
+    # Check values are correctly passed to 'deploy.deploy.deploy_contract'
```
martriay, line 148: `deploy.deploy.deploy` 😅

tests/commands/test_account.py
```diff
@@ -120,6 +122,43 @@ def test_declare(mock_declare, mock_get_class, mock_deploy):
...
+    mock_deploy_contract.assert_called_with(
+        account,
+        contract_name,
+        salt,
+        unique,
+        calldata,
+        alias,
+        deployer_address,
+        max_fee,
+        abi=None,
```
martriay, line 158: can `abi` be different? should we include it?

tests/commands/test_deploy.py
```diff
@@ -97,3 +107,96 @@ def test_deploy(
...
     )
     assert f"🧾 Transaction hash: {hex(TX_HASH)}" in caplog.text
+
+
+@pytest.mark.parametrize(
+    "args, exp_class_hash, exp_salt, exp_abi",
+    [
+        (
+            [CONTRACT, 0, True, [], ALIAS, 0x424242, 5],  # args
+            "0x434343",
```
martriay, line 117: maybe moving these to a `CLASS_HASH` constant (or array) for legibility? maybe also collapsing into multiple `parametrize()` statements, to get a better idea of the combinations

martriay (commented):
looking very good! left some comments

src/nile/signer.py
```diff
@@ -44,10 +44,10 @@ def sign_declare(self, sender, contract_class, nonce, max_fee):
 
         return self.sign(message_hash=transaction_hash)
 
-    def sign_transaction(
+    def sign_invoke_tx(
```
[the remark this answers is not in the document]
ericnordelo: I think `sing_transaction` is too generic in this case because we have multiple transactions types, and this method is signing invoke ones specifically (we have a `sign_declare_tx` in the same file, and we probably will add a `sign_deploy_account_tx` for the deployment of accounts from recent protocol updates). 

The invoke name comes from starknet naming, as you can see [here](https://github.com/starkware-libs/cairo-lang/blob/f951a669e264796d53bf7a3cfd23d5bc1f6e54b8/src/starkware/starknet/cli/starknet_cli.py#L55).

If we let `sign_transaction`, why shouldn't this method be used to sign declare transactions? or deploy_account transactions?

README.md
ericnordelo: My first thought was the same (to follow the same format of send), but as you state, the difference here is that we need to support both with and without account options. Click parameters (instead of options) are not optional (they could be, but with important restrictions I think are not worth to address). 

IMO the best approach would be to let account an option in this PR (because it is potentially the desired behavior anyway) and fix (or improve) pk handling later in the issues you mentioned.

README.md
ericnordelo: But we need to enforce at least an interface supporting the `deployContract` method as the deployer. If the deployer is not UDC compliant (not supporting this method) the deployment through the account won't work (at least in this version), wouldn't be good then to let implicit this requirement at least on the name of the option? (you can have a deployer extending UDC, but you need to support the UDC interface)

src/nile/cli.py
ericnordelo: I assumed that we shouldn't need a mainnet token to deploy through an invoke transaction as we are doing here, but I must have confirmed this with starkware team. Let me follow up on slack with them.

src/nile/core/account.py
ericnordelo: True. This can be considered external, so the comment is misleading, but this address must be of type int anyway to follow the hex int pattern rule:

- Always **return int from Internal or Public API, except when the method explicitly declares the intention to return a hex** (like `nile.utils.hex_class_hash` helper).

The rationale behind this rule is to avoid parsing these values often in the source code, where we use int internally (it is simpler and more organized to convert to hex when required instead of everywhere).

Note that this rule is not for outputting to files or CLI, it is for returning values of functions or properties.

I will remove the comment, but I would let the assertion to avoid misuse of the pattern (and I'm in favor of more assertions in the code).

tests/commands/test_account.py
ericnordelo: Fixed :)

src/nile/cli.py
ericnordelo: From the answer, i think I assumed correctly.

tests/commands/test_deploy.py
ericnordelo: I refactored for better readability (constants). Still, I don't like too much the idea of parametrizing the args because it would be adding seven more parameters at least to the `test_deploy_contract` method, which already has 9, and I don't think the benefit is big enough.

README.md
martriay: well the issue is that the standard defines the UDC as unique too. it has to be deployed to a specific address therefore any other deployer is not the UDC, let alone be compliant. i don't think it's bad to assume that such a deployer contract will follow this interface. I think we should create a standard interface for deployers too.

martriay: https://community.starknet.io/t/simp-deployer-contract-interface/2772

README.md
martriay: > the difference here is that we need to support both
i'm not sure about this. "accountless" deployments are already discouraged and will be prohibited soon. UDC-like deployments are already functional, so we don't really need to support it.

i think it's preferable to have a consistent, expectable API than an inconsistent one with perfect/final portions. i think we should aim for consistency and eventually migrate all methods to use `--account`.

martriay (commented):
new changes looking good!

README.md
ericnordelo: Are you suggesting then removing the support for deploys without accounts? If we remove this support, the current starknet-devnet version we support in Nile (pre 0.10.1) won't work for deployments because it has not predeployed UDC.

README.md
ericnordelo: That makes sense, I will rename it to `deployer_address` then.

README.md
martriay: yes that's the suggestion, since this release will require devnet@0.3.5 already. happy to keep the functionality as long as it doesn't compromise API consistency if there's such alternative

ericnordelo: I changed the API to match the `send` one, but let a `--ignore_account` flag to deploy without an account with the deprecation warning until the direct deploys are fully removed.

martriay: thanks. i asked the starkware team, they expect to remove it by this week.

src/nile/signer.py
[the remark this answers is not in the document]
martriay: agreed. then what do you think about renaming `sign_invoke_tx` -> `sign_execute` and dropping `_tx` from them all?

README.md
```diff
@@ -119,30 +119,35 @@ Creating artifacts/abis/ to store compilation artifacts
...
 ```txt
-nile deploy <contract> [--alias ALIAS] [--network NETWORK] [--track | --debug]
+nile deploy <private_key_alias> <contract> [--alias ALIAS] [--network NETWORK] [--track | --debug]
 ```
 
+For example:
+
 ```sh
-nile deploy contract --alias my_contract
+nile deploy setup_account_alias contract --alias my_contract
```
martriay, line 133:
```suggestion
nile deploy <private_key_alias> contract --alias my_contract
```
this is consistent with the rest of the readme

martriay (commented):
Thanks for the changes! This is almost done

src/nile/signer.py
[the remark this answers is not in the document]
andrew-fleming: +1 to removing the `_tx`

src/nile/core/deploy.py
```diff
@@ -47,3 +62,60 @@ def deploy(
...
+    )
+    register_abi = abi if abi is not None else f"{base_path[1]}/{contract_name}.json"
+    deployer_for_address_generation = 0
+
+    if unique:
+        # Match UDC address generation
+        salt = compute_hash_chain(data=[account.address, salt])
+        deployer_for_address_generation = deployer_address
+
+    # This needs to be fixed following the Hex/Int Pattern
```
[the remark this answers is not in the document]
ericnordelo: Fix for the future that I didn't implement because it had a dependency on another PR.

andrew-fleming (dismissed):
The improvements look great! Making the API consistent with `nile deploy <key> <contract>` is much more intuitive, so kudos on the change!

I left a `+1` to Marto's comment in removing the `_tx` on the signer methods. I don't think there's much to gain with it included.

Otherwise, this looks good to go for me :)

ericnordelo:
> I left a `+1` to Marto's comment in removing the `_tx` on the signer methods. I don't think there's much to gain with it included.

The _tx can be removed, but the change I wanted to introduce was replacing `sign_transaction` by `sign_invoke`, because sign_transaction sounded like it could be used to sign declares, or deploy_account, transaction, and this is not the case (_tx was because it looked better to me atm sign_invoke_tx than sign_invoke)

andrew-fleming:
> _tx was because it looked better to me atm sign_invoke_tx than sign_invoke

That's fair...`sign_invoke` doesn't look that great. I don't mind `sign_execute` though as previously mentioned, but no objections by me if we keep it as is

ericnordelo:
> That's fair...`sign_invoke` doesn't look that great. I don't mind `sign_execute` though as previously mentioned, but no objections by me if we keep it as is

I chose invoke over execute to follow the starknet source code wording (as you can see [here](https://github.com/starkware-libs/cairo-lang/blob/f951a669e264796d53bf7a3cfd23d5bc1f6e54b8/src/starkware/starknet/cli/starknet_cli.py#L55).)

martriay:
I get the point, but `Invoke` is a legacy term for when you could call any arbitrary entry point. I don't think following legacy terminology adds much clarity to users now you can only call `execute` with it.

ericnordelo:
> I get the point, but `Invoke` is a legacy term for when you could call any arbitrary entry point. I don't think following legacy terminology adds much clarity to users now that invoke is just a proxy to call `execute` with it.

That sounds fair, but we are still using invoke (not execute) in important parts of the project, like in logging for cli (with `nile send` for example, and other commands), in the README as well, or in the source code where we have `call_or_invoke`, and invoke type for transactions as some examples. I think `sign_invoke` is more consistent with what we have, but we could refactor this in a separate issue if you agree.

martriay (approved):
either `sign_invoke` or `sing_execute` are fine with me. as long as we drop `_tx` we can merge this now :)

ericnordelo:
> either `sign_invoke` or `sing_execute` are fine with me as long as we drop `_tx`. this is ready :)

Removed the _tx suffix :)

martriay (approved):

andrew-fleming (approved):
I think we're good to go!

Issue #287, opened 2022-11-14
account.declare creates duplicate entries in declarations file if called multiple times


ericglau:
Using `main` branch, when you run `declare` multiple times e.g.:
```
nile declare PKEY1 contract
nile declare PKEY1 contract
```
This creates duplicate entries in `localhost.declarations.txt` e.g.:
```
0x0f28592e02fb10ae73e235b5959bd18297efaa4333ded153bdbefc785a243ef
0x0f28592e02fb10ae73e235b5959bd18297efaa4333ded153bdbefc785a243ef
```

ericglau:
I think the issue is as follows:
`declare` passes a hex class hash (string) to `deployments.register_class_hash` -> `class_hash_exists` -> `load_class`.

`load_class` currently expects an int class hash, otherwise assumes the `hash_or_alias` parameter is an alias.

ericglau:
A related question: `account.declare` simply declares a class (submits a tx), then registers it in the declarations file afterwards.

Before doing the declare, should it check the declarations file and/or check on-chain to see if the class has previously been declared?  And if it was previously declared, should it fail or succeed (as a no-op)?

PR #288, opened 2022-11-14
Change `register_class_hash` to take int

ericglau:
Fixes https://github.com/OpenZeppelin/nile/issues/287

According to [hex/int pattern](https://github.com/OpenZeppelin/nile/blob/main/CONTRIBUTING.md#hexint-pattern), internal API must use only hex.

I assume https://github.com/OpenZeppelin/nile/blob/main/src/nile/deployments.py is considered internal API (since it is only intended for use by Nile and plugins).

So functions relating to class hashes in the deployment files should take `int` but write to log and files as hex strings.

### DIFF
--- a/src/nile/core/declare.py
+++ b/src/nile/core/declare.py
@@ -44,7 +44,7 @@ def declare(
     logging.info(f"⏳ Successfully sent declaration of {contract_name} as {padded_hash}")
     logging.info(f"🧾 Transaction hash: {hex(tx_hash)}")
 
-    deployments.register_class_hash(padded_hash, network, alias)
+    deployments.register_class_hash(class_hash, network, alias)
 
     if watch_mode is not None:
         if status(tx_hash, network, watch_mode).status.is_rejected:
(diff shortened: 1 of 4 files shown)


ericglau:
A related question: `account.declare` simply declares a class (submits a tx), then registers it in the declarations file afterwards.

Before doing the declare, it checks if the declarations file already has an identical alias, but does not check for an identical class hash.

Should it check the declarations file and/or check on-chain to see if a class has previously been declared?  And if it was previously declared, should it fail or succeed (as a no-op)?

ericnordelo (approved):
Looking good Eric! Good catch.

ericnordelo:
> Before doing the declare, it checks if the declarations file already has an identical alias, but does not check for an identical class hash.
> 
> Should it check the declarations file and/or check on-chain to see if a class has previously been declared? And if it was previously declared, should it fail or succeed (as a no-op)?

We plan to remove the declaration file (@martriay and @andrew-fleming can confirm this). There is no real value in registering a declaration in Nile (besides executing the declare transaction itself) because declarations are not used from off-chain besides deployments. We can get the class_hash from the contract code anyway. The other use is for library_calls, but this is from Cairo code. No need for off-chain registrations.

Issue #301, opened 2022-11-23
Account invariant can be broken


martriay:
When deploying an account contract to construct an Account instance on Nile, the returned address can be a string thus violating the `int` invariant:

https://github.com/OpenZeppelin/nile/blob/b81dce88bb07582dd207bebe8d8412527cc1334b/src/nile/core/account.py#L94-L104

PR #302, opened 2022-11-23
Fix account invariant

martriay:
Fixes https://github.com/OpenZeppelin/nile/issues/301. This PR also fixes `max_fee`, which was being passed to `starknet_cli` as an array instead of a number:

`['1', '2', '3', '4']` instead of `1234`

### DIFF
--- a/src/nile/core/account.py
+++ b/src/nile/core/account.py
@@ -96,7 +96,7 @@ class Account(AsyncObject):
             )
             if output is not None:
                 address, index = output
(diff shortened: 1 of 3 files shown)


andrew-fleming (approved):
Good catch!

martriay:
i think we should release this as a patch asap cc/ @ericnordelo

ericnordelo (approved):
Looking good!

PR #303, opened 2022-11-23
Fix network handling

### LINKED ISSUE #297: Issue trying to use goerli2 in commands
**Running**

`nile setup account --network goerli2`

**Receiving error:**

```
🚀 Deploying Account
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/bin/nile", line 8, in <module>
    sys.exit(cli())
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/asyncclick/core.py", line 1157, in __call__
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/anyio/_core/_eventloop.py", line 70, in run
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 292, in run
  File "/Users/ericnordelo/.pyenv/versions/3.9.1/lib/python3.9/asyncio/runners.py", line 44, in run
  File "/Users/ericnordelo/.pyenv/versions/3.9.1/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 287, in wrapper
    return await func(*args)
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/asyncclick/core.py", line 1160, in _main
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/asyncclick/core.py", line 1076, in main
    rv = await self.invoke(ctx)
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/asyncclick/core.py", line 1687, in invoke
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/asyncclick/core.py", line 1434, in invoke
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/asyncclick/core.py", line 780, in invoke
    rv = await rv
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/nile/cli.py", line 152, in setup
    await Account(signer, network, watch_mode=watch_mode)
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/nile/core/account.py", line 35, in __new__
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/nile/core/account.py", line 81, in __init__
    address, index = await self.deploy(watch_mode=watch_mode)
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/nile/core/account.py", line 91, in deploy
    address, _ = await deploy(
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/nile/core/deploy.py", line 29, in deploy
    output = await execute_call(
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/nile/starknet_cli.py", line 29, in execute_call
    return await capture_stdout(cmd(args=args, command_args=command_args))
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/nile/starknet_cli.py", line 36, in capture_stdout
    await func
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/starkware/starknet/cli/starknet_cli.py", line 846, in deploy
    await deploy_tx(args=args)
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/starkware/starknet/cli/starknet_cli.py", line 884, in deploy_tx
    gateway_client = get_gateway_client(args)
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/starkware/starknet/cli/starknet_cli.py", line 202, in get_gateway_client
    gateway_url = get_arg_value(
  File "/Users/ericnordelo/Documents/OpenSource/OpenZeppelin/nile/env/lib/python3.9/site-packages/starkware/starknet/cli/starknet_cli.py", line 147, in get_arg_value
    f'{arg_name} must be specified with the "{args.command}" subcommand.\n'
AttributeError: 'types.SimpleNamespace' object has no attribute 'command'
```

**Expected behavior:**

Correct deployment of account.

**How to reproduce:**

Install from main branch with `pip install .`

Run `nile setup account --network goerli2`

**Notes**

This happens also with other commands like `nile deploy a UniversalDeployer --alias udc --ignore_account --network goerli2`

andrew-fleming:
Resolves #297.

The current flow of network handling creates the awkward issue of an unresolved query in perpetuity when the node.json exists but the network key does not. This PR, therefore, proposes to add `goerli2` and `integration` gateways to the node.json in the `nile node` command. This means that the `goerli2` and `integration` gateways will be accessible irrespective of the node.json file's existence.

### DIFF
--- a/src/nile/common.py
+++ b/src/nile/common.py
@@ -29,6 +29,7 @@ UNIVERSAL_DEPLOYER_ADDRESS = (
     "0x041a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf"
 )
 DEFAULT_GATEWAYS = {
+    "localhost": "http://127.0.0.1:5050/",
     "goerli2": "https://alpha4-2.starknet.io",
     "integration": "https://external.integration.starknet.io",
 }
@@ -36,23 +37,27 @@ DEFAULT_GATEWAYS = {
 
 def get_gateways():
     """Get the StarkNet node details."""
-    try:
+    if os.path.exists(NODE_FILENAME):
         with open(NODE_FILENAME, "r") as f:
             gateway = json.load(f)
-            gateways = {**gateway, **DEFAULT_GATEWAYS}
+            gateways = {**DEFAULT_GATEWAYS, **gateway}
             return gateways
-    except FileNotFoundError:
-        gateway = create_node_json()
-        gateways = {**gateway, **DEFAULT_GATEWAYS}
-        return gateways
+    else:
+        return DEFAULT_GATEWAYS
 
 
-def create_node_json(network="localhost", gateway_url="http://127.0.0.1:5050/"):
-    """Create node.json and return gateways."""
-    with open(NODE_FILENAME, "w") as f:
-        gateway = {network: gateway_url}
-        f.write(json.dumps(gateway))
-        return gateway
+def create_node_json(network, gateway_url):
+    """Create or update node.json with custom network."""
+    if not os.path.exists(NODE_FILENAME):
+        with open(NODE_FILENAME, "w") as fp:
+            json.dump({network: gateway_url}, fp)
+    else:
+        with open(NODE_FILENAME, "r") as fp:
+            gateways = json.load(fp)
+            gateways[network] = gateway_url
+
+        with open(NODE_FILENAME, "w") as fp:
+            json.dump(gateways, fp, indent=2)
 
 
 GATEWAYS = get_gateways()
--- a/src/nile/common.py
+++ b/src/nile/common.py
@@ -28,7 +28,7 @@ UNIVERSAL_DEPLOYER_ADDRESS = (
     # subject to change
     "0x041a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf"
 )
-OTHER_NETWORKS = {
+DEFAULT_GATEWAYS = {
     "goerli2": "https://alpha4-2.starknet.io",
     "integration": "https://external.integration.starknet.io",
 }
@@ -43,10 +43,10 @@ def get_gateways():
 
     except FileNotFoundError:
         with open(NODE_FILENAME, "w") as f:
-            networks = {"localhost": "http://127.0.0.1:5050/", **OTHER_NETWORKS}
-            f.write(json.dumps(networks, indent=2))
+            gateways = {"localhost": "http://127.0.0.1:5050/", **DEFAULT_GATEWAYS}
+            f.write(json.dumps(gateways, indent=2))
 
-            return networks
+            return gateways
 
 
 GATEWAYS = get_gateways()
--- a/src/nile/common.py
+++ b/src/nile/common.py
@@ -38,20 +38,21 @@ def get_gateways():
     """Get the StarkNet node details."""
     try:
         with open(NODE_FILENAME, "r") as f:
-            gateways = json.load(f)
+            gateway = json.load(f)
+            gateways = {**gateway, **DEFAULT_GATEWAYS}
             return gateways
-
     except FileNotFoundError:
-        gateways = create_node_json()
+        gateway = create_node_json()
+        gateways = {**gateway, **DEFAULT_GATEWAYS}
         return gateways
 
 
 def create_node_json(network="localhost", gateway_url="http://127.0.0.1:5050/"):
     """Create node.json and return gateways."""
     with open(NODE_FILENAME, "w") as f:
-        gateways = {network: gateway_url, **DEFAULT_GATEWAYS}
-        f.write(json.dumps(gateways, indent=2))
-        return gateways
+        gateway = {network: gateway_url}
+        f.write(json.dumps(gateway))
+        return gateway
 
 
 GATEWAYS = get_gateways()
--- a/src/nile/common.py
+++ b/src/nile/common.py
@@ -28,6 +28,10 @@ UNIVERSAL_DEPLOYER_ADDRESS = (
     # subject to change
     "0x041a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf"
 )
+OTHER_NETWORKS = {
+    "goerli2": "https://alpha4-2.starknet.io",
+    "integration": "https://external.integration.starknet.io",
+}
 
 
 def get_gateways():
@@ -39,11 +43,7 @@ def get_gateways():
 
     except FileNotFoundError:
         with open(NODE_FILENAME, "w") as f:
-            networks = {
-                "localhost": "http://127.0.0.1:5050/",
-                "goerli2": "https://alpha4-2.starknet.io",
-                "integration": "https://external.integration.starknet.io",
-            }
+            networks = {"localhost": "http://127.0.0.1:5050/", **OTHER_NETWORKS}
             f.write(json.dumps(networks, indent=2))
 
             return networks
--- a/src/nile/common.py
+++ b/src/nile/common.py
@@ -39,8 +39,8 @@ def get_gateways():
     """Get the StarkNet node details."""
     if os.path.exists(NODE_FILENAME):
         with open(NODE_FILENAME, "r") as f:
-            gateway = json.load(f)
-            gateways = {**DEFAULT_GATEWAYS, **gateway}
+            custom_gateways = json.load(f)
+            gateways = {**DEFAULT_GATEWAYS, **custom_gateways}
             return gateways
     else:
         return DEFAULT_GATEWAYS
--- a/src/nile/common.py
+++ b/src/nile/common.py
@@ -46,7 +46,7 @@ def get_gateways():
         return DEFAULT_GATEWAYS
 
 
-def create_node_json(network, gateway_url):
+def write_node_json(network, gateway_url):
     """Create or update node.json with custom network."""
     if not os.path.exists(NODE_FILENAME):
         with open(NODE_FILENAME, "w") as fp:
--- a/src/nile/common.py
+++ b/src/nile/common.py
@@ -52,11 +52,10 @@ def write_node_json(network, gateway_url):
         with open(NODE_FILENAME, "w") as fp:
             json.dump({network: gateway_url}, fp)
     else:
-        with open(NODE_FILENAME, "r") as fp:
+        with open(NODE_FILENAME, "r+") as fp:
             gateways = json.load(fp)
             gateways[network] = gateway_url
-
-        with open(NODE_FILENAME, "w") as fp:
+            fp.seek(0)
             json.dump(gateways, fp, indent=2)
 
 
--- a/src/nile/common.py
+++ b/src/nile/common.py
@@ -38,15 +38,20 @@ def get_gateways():
     """Get the StarkNet node details."""
     try:
         with open(NODE_FILENAME, "r") as f:
-            gateway = json.load(f)
-            return gateway
+            gateways = json.load(f)
+            return gateways
 
     except FileNotFoundError:
-        with open(NODE_FILENAME, "w") as f:
-            gateways = {"localhost": "http://127.0.0.1:5050/", **DEFAULT_GATEWAYS}
-            f.write(json.dumps(gateways, indent=2))
+        gateways = create_node_json()
+        return gateways
 
-            return gateways
+
+def create_node_json(network="localhost", gateway_url="http://127.0.0.1:5050/"):
+    """Create node.json and return gateways."""
+    with open(NODE_FILENAME, "w") as f:
+        gateways = {network: gateway_url, **DEFAULT_GATEWAYS}
+        f.write(json.dumps(gateways, indent=2))
+        return gateways
 
 
 GATEWAYS = get_gateways()
--- a/src/nile/core/node.py
+++ b/src/nile/core/node.py
@@ -2,7 +2,7 @@
 import logging
 import subprocess
 
-from nile.common import DEFAULT_GATEWAYS, create_node_json
+from nile.common import DEFAULT_GATEWAYS, write_node_json
 
 
 def node(host="127.0.0.1", port=5050, seed=None, lite_mode=False):
@@ -15,7 +15,7 @@ def node(host="127.0.0.1", port=5050, seed=None, lite_mode=False):
             network = host
         gateway_url = f"http://{host}:{port}/"
         if DEFAULT_GATEWAYS.get(network) != gateway_url:
-            create_node_json(network, gateway_url)
+            write_node_json(network, gateway_url)
 
         command = ["starknet-devnet", "--host", host, "--port", str(port)]
 
--- a/src/nile/core/node.py
+++ b/src/nile/core/node.py
@@ -3,7 +3,7 @@ import json
 import logging
 import subprocess
 
-from nile.common import NODE_FILENAME, OTHER_NETWORKS
+from nile.common import NODE_FILENAME, DEFAULT_GATEWAYS
 
 
 def node(host="127.0.0.1", port=5050, seed=None, lite_mode=False):
@@ -16,7 +16,7 @@ def node(host="127.0.0.1", port=5050, seed=None, lite_mode=False):
         else:
             network = host
         gateway_url = f"http://{host}:{port}/"
-        gateways = {network: gateway_url, **OTHER_NETWORKS}
+        gateways = {network: gateway_url, **DEFAULT_GATEWAYS}
 
         with open(file, "w+") as f:
             json.dump(gateways, f, indent=2)
--- a/src/nile/core/node.py
+++ b/src/nile/core/node.py
@@ -2,7 +2,7 @@
 import logging
 import subprocess
 
-from nile.common import create_node_json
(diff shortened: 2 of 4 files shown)


martriay (dismissed):
good patch. but i think the real solution is to centralize file writes to avoid this kind of bug. it's also a bit awkward that `get_gateways` writes anywhere, we should either refactor or rename.

andrew-fleming:
Agreed on centralizing file writes and with that, a refactor on `get_gateways`. Should we keep this patch for now or go for the refactor in this PR?

martriay:
if you think it's small enough, let's squeeze it here. otherwise the patch + tracking the issue is enough.

andrew-fleming:
> if you think it's small enough, let's squeeze it here.

On it!

src/nile/common.py
```diff
@@ -28,6 +28,10 @@
     # subject to change
     "0x041a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf"
 )
+OTHER_NETWORKS = {
```
ericnordelo, line 31:
```suggestion
DEFAULT_GATEWAYS = {
```

I think this is more consistent.

src/nile/common.py
```diff
@@ -39,11 +43,7 @@ def get_gateways():
 
     except FileNotFoundError:
         with open(NODE_FILENAME, "w") as f:
-            networks = {
```
ericnordelo, line 42: In the try block above, gateway should be in plural.

Also, I think this is still not working for sending transactions if goerli2 is not in the node.json file, but the file exists. OTHER_NETWORKS are added just when the file is not found. Maybe return it in the try block as well?

src/nile/common.py
```diff
@@ -39,11 +43,7 @@ def get_gateways():
...
-                "localhost": "http://127.0.0.1:5050/",
-                "goerli2": "https://alpha4-2.starknet.io",
-                "integration": "https://external.integration.starknet.io",
-            }
+            networks = {"localhost": "http://127.0.0.1:5050/", **OTHER_NETWORKS}
```
ericnordelo, line 46:
```suggestion
            gateways = {"localhost": "http://127.0.0.1:5050/", **DEFAULT_GATEWAYS}
```

src/nile/common.py
```diff
@@ -39,11 +43,7 @@ def get_gateways():
...
             f.write(json.dumps(networks, indent=2))
 
             return networks
```
ericnordelo, line 49:
```suggestion
            return gateways
```

src/nile/common.py
```diff
@@ -39,11 +43,7 @@ def get_gateways():
 
     except FileNotFoundError:
         with open(NODE_FILENAME, "w") as f:
-            networks = {
-                "localhost": "http://127.0.0.1:5050/",
-                "goerli2": "https://alpha4-2.starknet.io",
-                "integration": "https://external.integration.starknet.io",
-            }
+            networks = {"localhost": "http://127.0.0.1:5050/", **OTHER_NETWORKS}
             f.write(json.dumps(networks, indent=2))
```
ericnordelo, line 47:
```suggestion
            f.write(json.dumps(gateways, indent=2))
```

src/nile/core/node.py
```diff
@@ -3,7 +3,7 @@
 import logging
 import subprocess
 
-from nile.common import NODE_FILENAME
+from nile.common import NODE_FILENAME, OTHER_NETWORKS
```
ericnordelo, line 6:
```suggestion
from nile.common import NODE_FILENAME, DEFAULT_GATEWAYS
```

src/nile/core/node.py
```diff
@@ -16,10 +16,10 @@ def node(host="127.0.0.1", port=5050, seed=None, lite_mode=False):
         else:
             network = host
         gateway_url = f"http://{host}:{port}/"
-        gateway = {network: gateway_url}
+        gateways = {network: gateway_url, **OTHER_NETWORKS}
```
ericnordelo, line 19:
```suggestion
        gateways = {network: gateway_url, **DEFAULT_GATEWAYS}
```

tests/test_common.py
```diff
@@ -1,7 +1,13 @@
 """Tests for common library."""
 import pytest
 
-from nile.common import parse_information, prepare_params, stringify
+from nile.common import (
+    OTHER_NETWORKS,
```
ericnordelo, line 5:
```suggestion
    DEFAULT_GATEWAYS,
```

tests/test_common.py
```diff
@@ -48,3 +54,10 @@ def test_parse_information():
 
     _a, _b = parse_information(target)
     assert _a, _b == (a, b)
+
+
+def test_get_gateways():
+    gateways = get_gateways()
+    expected = {"localhost": "http://127.0.0.1:5050/", **OTHER_NETWORKS}
```
ericnordelo, line 61:
```suggestion
    expected = {"localhost": "http://127.0.0.1:5050/", **DEFAULT_GATEWAYS}
```

src/nile/common.py
ericnordelo: If we do this, I think it makes sense to remove writing the DEFAULT_NETWORKS to the node.json file at all, because it will be part of GATEWAYS always anyway.

src/nile/common.py
ericnordelo, line 46: If we add DEFAULT_GATEWAYS to queries even when the node.json exists:

```suggestion
            networks = {"localhost": "http://127.0.0.1:5050/"}
```

ericnordelo (commented):
Thanks for the fix @andrew-fleming! Left some comments.

src/nile/common.py
andrew-fleming: Good idea!

andrew-fleming: > I think it makes sense to remove writing the DEFAULT_NETWORKS to the node.json file

It seems difficult to handle (at least, to cleanly handle) `localhost` network calls when the node.json is not present with this approach^. I think we'd have to add conditionals to `get_gateway_url` and `get_feeder_url` in starknet_cli.py. Maybe there's a better way. If it's agreeable, I'd say let's make another issue to track that for the future.

ericnordelo (commented):
@andrew-fleming The refactor looks good, but this PR is not solving the linked issue imo. Even when the error message differs from the last merges, the expected behavior, for me at least, is goerli2 transactions working by default (there are no docs telling users they need to add this manually to node.json afaik). 

I think we have enough content with these different networks to add a `nile.ini` (or different name) required config file with networks instead of deleting/creating a `node.json` file.

Adding these networks by default to the ini file and documenting this would be a huge improvement to the UX imo.

src/nile/common.py
ericnordelo: Why do we need to handle localhost calls when the node.json is not present with this approach? The suggestion is to remove DEFAULT_NETWORKS, which are goerli2 and integration, which gateway_urls won't probably change often anyway. localhost calls would still be going through node.json. Sorry if I'm missing something.

andrew-fleming: Forgive me, I did not provide enough context in my response.

The idea was to handle `localhost` separately from `DEFAULT_NETWORKS` through some sort of conditional. Otherwise, a node.json will be created (if it's not already) for txs querying the `DEFAULT_NETWORKS`. This doesn't break anything; however, it's not ideal. This occurs because `GATEWAYS.get()` returns None when a key does not exist instead of an error that we can handle.

I'm interested in the networks config idea.

src/nile/core/node.py
```diff
@@ -1,25 +1,20 @@
 """Command to start StarkNet local network."""
-import json
 import logging
 import subprocess
 
-from nile.common import NODE_FILENAME, DEFAULT_GATEWAYS
+from nile.common import create_node_json
 
 
 def node(host="127.0.0.1", port=5050, seed=None, lite_mode=False):
```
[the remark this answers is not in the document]
andrew-fleming: > Having two local networks configured shouldn't be allowed?

IMO it's not necessary. Executing `nile node` while a local node is already running would indeed overwrite the dictionary. We can add a check to prevent that. If we want to allow multiple local nodes, we'll need to give a deeper refactor on network handling

andrew-fleming:
> this PR is not solving the linked issue imo. Even when the error message differs from the last merges, the expected behavior, for me at least, is goerli2 transactions working by default (there are no docs telling users they need to add this manually to node.json afaik).

The patch does solve that^. The user doesn't have to add anything manually. The flow was that if a node.json exists, `localhost` and `DEFAULT_NETWORKS` would already be written in the json. If the node.json doesn't exist, it would be created with all three networks ready to go. Please correct me if I'm misunderstanding the comment.

This PR now (since after that review) doesn't write the `DEFAULT_NETWORKS` to the node.json. Instead, it follows the suggestion of returning the contents of node.json (which consists only of `localhost`) with the `DEFAULT_GATEWAYS` dictionary from common.py.

> 1. Issue resolution that I think could be solved by returning the DEFAULT_GATEWAYS in the get_gateways method even when they are not in the node.json file.
> 2. Replace node.json with nile.ini file with a networks config.

The latest pushed changes follow `#1`. The one minor issue with this flow is that a node.json will be created (if it doesn't already exist) when querying `DEFAULT_NETWORKS`.

If we want to go for `#2`, we should adjust the issue size or create a new issue :)

src/nile/common.py
ericnordelo: Got it. 

My issue with the current PR state is that it is not addressing the real issue IMO, which is the UX using Nile. We can use a different approach than the one I suggested in this thread (mine is probably not the best for sure), but letting this as it is, I fear users will upgrade Nile, and they will struggle with sending transactions to goerli2 or integration. You and I both know how to fix it, but I feel the error they will receive is not well documented (and I think we could prevent the issue by default). Is the details what makes great tooling IMO (hardhat vs truffle).

src/nile/core/node.py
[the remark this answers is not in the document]
ericnordelo: I'm not worried about overwriting the localhost config, but this is overwriting the whole file.

What if I had **mainnet-fork** network configured? (this would be possible soon), but I will lose it by just running `nile node`.

I think a `nile.ini` file is a stronger solution for this.

ericnordelo:
> The patch does solve that^. The user doesn't have to add anything manually. The flow was that if a node.json exists, `localhost` and `DEFAULT_NETWORKS` would already be written in the json. If the node.json doesn't exist, it would be created with all three networks ready to go. Please correct me if I misunderstood the comment.

This is assuming users would follow the flow, but my concern was also with upgrades, like what happened to me: Dev already have a nile project, already having a node.json file, they try to use goerli2, cryptic error not documented.

> If we want to go for #2, we should adjust the issue size or create a new issue :)

Totally agree. I think we should prioritize this as possible (not necessarily my proposed solution, but something to improve the network handling). cc @martriay

src/nile/common.py
martriay: this is my middle ground suggestion:

- let's not write any node.json. it's confusing pollution to the user, and it's information nile already knows. no more writing conflicts nor retrocompatibility issues.
- let's still read node.json. if there's any user custom network, let's add it (or overwrite any default). this is the stem for an actual nile config file, but retrocompatible and it buys us time to design properly, while fixing this.

thougths?

ericnordelo:
> this is my middle ground suggestion:
>
> let's not write any node.json. it's confusing pollution to the user, and it's information nile already knows. no more writing conflicts nor retrocompatibility issues.
>
> let's still read node.json. if there's any user custom network, let's add it (or overwrite any default). this is the stem for an actual nile config file, but retrocompatible and it buys us time to design properly, while fixing this.
>
> thoughts?

Just confirming. You mean not writing to the node.json file after creating it, but keeping all network configurations and reading from it, right? I'd keep all the networks in a network config, to allow changing the RPC providers (gateway_urls) if we have more than one at some point for the same network (I assume we'll have).

src/nile/common.py
```diff
@@ -28,25 +28,31 @@
...
 
 
 def get_gateways():
     """Get the StarkNet node details."""
     try:
         with open(NODE_FILENAME, "r") as f:
             gateway = json.load(f)
-            return gateway
-
+            gateways = {**gateway, **DEFAULT_GATEWAYS}
```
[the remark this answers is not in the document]
andrew-fleming: If there's a conflicting network, the `DEFAULT_GATEWAYS` network would overwrite it in the local variable `gateways` when the gateways dict is passed to the querying function. Thus, `"goerli2": "new_url"` in node.json will return the goerli2 from `DEFAULT_NETWORKS`

I'll switch the order so that the node.json's gateway overwrites the `DEFAULT_GATEWAYS` network of the same name

martriay:
> Just confirming. You mean not writing to the node.json file after creating it, but keeping all network configurations and reading from it, right? I'd keep all the networks in a network config, to allow changing the RPC providers (gateway_urls) if we have more than one at some point for the same network (I assume we'll have).

I mean not write/create any node.json files anymore, but take it into account if there's one already (e.g. with a custom network). this would be retrocompatible and fix the UX issue you point out, without us introducing any weird logic to overwrite/fix existing node.json files, and also provide a way for users to add their custom ones.

node.json would be the network file you suggest, my approach only simplifies it so we can ship a fix asap. this node.json file can eventually evolve into a full config file #73, but i'm leaving it for later to avoid the design + refactor time it entails.

andrew-fleming:
Here's a few notes on the latest changes aiming to address the conversations regarding this PR @martriay @ericnordelo:
- The default `localhost` now belongs in `DEFAULT_GATEWAYS`.
- node.json will only be written when `nile node` receives a different host or port i.e. `nile node --port 5001`.
  - In the event that the `localhost` key points to a different value (this key/value entry would only exist in the node.json), the network will favor the node.json value.
  - When the node.json receives new entries, the previous entries will not be deleted. Should the node.json receive a duplicate key, however, the new value _will_ overwrite the old.

src/nile/common.py
```diff
@@ -28,25 +28,36 @@
...
+    if not os.path.exists(NODE_FILENAME):
+        with open(NODE_FILENAME, "w") as fp:
+            json.dump({network: gateway_url}, fp)
+    else:
+        with open(NODE_FILENAME, "r") as fp:
+            gateways = json.load(fp)
+            gateways[network] = gateway_url
+
+        with open(NODE_FILENAME, "w") as fp:
+            json.dump(gateways, fp, indent=2)
```
martriay, line 60:
```suggestion
        with open(NODE_FILENAME, "w") as fp:
            gateways = json.load(fp)
            gateways[network] = gateway_url
            json.dump(gateways, fp, indent=2)
```
isn't this simpler? why indent is only explicit here?

src/nile/common.py
```diff
@@ -28,25 +28,36 @@
...
-            f.write(json.dumps(networks, indent=2))
-
-            return networks
+            gateways = {**DEFAULT_GATEWAYS, **gateway}
+            return gateways
+    else:
+        return DEFAULT_GATEWAYS
+
+
+def create_node_json(network, gateway_url):
```
martriay, line 49:
```suggestion
def write_node_json(network, gateway_url):
```
it only creates if it doesn't exist

martriay (approved):

src/nile/common.py
andrew-fleming: This is simpler^, but the `w` will remove the prior contents. This gives me an idea though

martriay (dismissed):

src/nile/common.py
```diff
@@ -28,25 +28,35 @@
...
+    "integration": "https://external.integration.starknet.io",
+}
 
 
 def get_gateways():
     """Get the StarkNet node details."""
-    try:
+    if os.path.exists(NODE_FILENAME):
         with open(NODE_FILENAME, "r") as f:
             gateway = json.load(f)
```
ericnordelo, line 42:
```suggestion
            custom_gateways = json.load(f)
```

The name is a bit confusing being possible to have multiple custom gateways in the `node.json` file.

ericnordelo (approved):
Looking good! Left just a small suggestion and I think we are good to go.

ericnordelo (approved):
LGTM!

PR #311, opened 2022-11-28
Return address, tx_hash from Account.deploy_contract

### LINKED ISSUE #307: `Account.deploy_contract` should return with tx hash and/or address
`Account.deploy_contract` calls https://github.com/OpenZeppelin/nile/blob/98ce378842eb5d404f82e3bc2d082dd06091b665/src/nile/core/account.py#L198
but omits the return value.  It would be useful for the caller if this returns information such as tx hash and/or address.

ericnordelo:
Fixes #307

### DIFF
--- a/src/nile/core/account.py
+++ b/src/nile/core/account.py
@@ -137,7 +137,7 @@ class Account(AsyncObject):
         )
 
         if output is not None:
-            address, _ = output
+            address, *_ = output
             accounts.register(
                 self.signer.public_key, address, index, self.alias, self.network
             )
--- a/src/nile/core/account.py
+++ b/src/nile/core/account.py
@@ -195,7 +195,7 @@ class Account(AsyncObject):
             deployer_address or UNIVERSAL_DEPLOYER_ADDRESS
         )
 
-        await deploy_with_deployer(
+        return await deploy_with_deployer(
             self,
             contract_name,
             salt,
--- a/tests/commands/test_account.py
+++ b/tests/commands/test_account.py
@@ -19,6 +19,8 @@ NETWORK = "localhost"
 MOCK_ADDRESS = 0x123
 MOCK_TARGET_ADDRESS = 0x987
 MOCK_INDEX = 0
+MOCK_ABI = "MOCK_ABI"
+MOCK_TX_HASH = 1
 MAX_FEE = 10
 SALT = 444
 SIGNATURE = [111, 222]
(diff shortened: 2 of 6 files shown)


andrew-fleming (commented):
Very nice, @ericnordelo! We should probably update the return value in the mocked `deploy_contract` as well:

https://github.com/OpenZeppelin/nile/blob/main/tests/commands/test_account.py#L152

Otherwise, this looks good to me!

ericglau (approved):

martriay (approved):

martriay:
looks good! addressing Andrew's comment should be enough

andrew-fleming (dismissed):
After a second look, I don't think the account tests need to be updated because the deploy mocks returning values only include the other deploys (`sign_deploy`, `deploy`, and `deploy_account`) and not `deploy_with_deployer` in test_account.py.

test_deploy.py and test_declare.py are updated, so this looks good to go for me!

martriay:
but [`deploy account`](https://github.com/OpenZeppelin/nile/pull/311/files#diff-c19a6bc8b864ba31b67d13c034aa2eba8da5047cfa2683bb708ce6773fbfdf21R187) did change its return value

src/nile/core/account.py
```diff
@@ -195,7 +195,7 @@ async def deploy_contract(
             deployer_address or UNIVERSAL_DEPLOYER_ADDRESS
         )
 
-        await deploy_with_deployer(
+        return await deploy_with_deployer(
```
martriay, line 198: this should be tested too

tests/commands/test_account.py
```diff
@@ -320,7 +349,10 @@ async def test_send_defaults(
 
 
 @pytest.mark.asyncio
-@patch("nile.core.account.deploy_account", return_value=(MOCK_ADDRESS, MOCK_INDEX))
+@patch(
+    "nile.core.account.deploy_account",
+    return_value=(MOCK_ADDRESS, MOCK_TX_HASH, MOCK_ABI),
```
martriay, line 354: it looks like we could abstract the whole tuple into a variable, to simplify these changes in the future

martriay (approved):

tests/commands/test_account.py
ericnordelo: Done!

martriay (approved):

PR #314, opened 2022-11-28
Remove watch_mode from NRE

### LINKED ISSUE #308: AssertionError `type(self.address) == int` when deploying Account
## Summary 
From a script:
- if `await Account(signer, nre.network, watch_mode="track")` is called, the account does not seem to get created.
- if `await Account(signer, nre.network)` is called, there is an error the first time, but it succeeds the second time.

## Sample project
https://github.com/ericglau/nile-account-test

## To reproduce
1. Run `nile init` to create a new project with Nile v0.11
2. Start a node with `nile node`
3. Create a script with
```
from nile.core.account import Account

async def run(nre):
    signer = "PRIVATE_KEY"
    await Account(signer, nre.network, watch_mode="track")
```
4. Run the script with `nile run <script>`.  Script returns but `localhost.accounts.json` and `localhost.deployments.txt` are empty.
5. Remove `, watch_mode="track"` from the script
6. Run the script with `nile run <script>`.  Script fails with 
```
  File "/Users/eric/git/nileproj3/env/bin/nile", line 8, in <module>
    sys.exit(cli())
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1157, in __call__
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/anyio/_core/_eventloop.py", line 70, in run
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 292, in run
  File "/opt/homebrew/Cellar/python@3.9/3.9.15/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/runners.py", line 44, in run
  File "/opt/homebrew/Cellar/python@3.9/3.9.15/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 287, in wrapper
    return await func(*args)
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1160, in _main
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1076, in main
    rv = await self.invoke(ctx)
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1687, in invoke
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1434, in invoke
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 780, in invoke
    rv = await rv
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/nile/cli.py", line 94, in run
    await run_command(path, network)
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/nile/core/run.py", line 14, in run
    await script.run(nre)
  File "scripts/run.py", line 6, in run
    await Account(signer, nre.network)
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/nile/core/account.py", line 40, in __new__
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/nile/core/account.py", line 104, in __init__
AssertionError
```
7. Run the script again.  Script passes and `localhost.accounts.json` and `localhost.deployments.txt` have the account.

ericnordelo:
Fixes #308 

Check [this comment](https://github.com/OpenZeppelin/nile/issues/308#issuecomment-1329505315) for context.

### DIFF
--- a/src/nile/nre.py
+++ b/src/nile/nre.py
@@ -33,7 +33,6 @@ class NileRuntimeEnvironment:
         overriding_path=None,
         abi=None,
         mainnet_token=None,
-        watch_mode=None,
     ):
         """Deploy a smart contract."""
         return deploy(
@@ -44,7 +43,6 @@ class NileRuntimeEnvironment:
             overriding_path=overriding_path,
             abi=abi,
             mainnet_token=mainnet_token,
-            watch_mode=watch_mode,
         )
 
     def call(self, address_or_alias, method, params=None, abi=None):
@@ -67,9 +65,9 @@ class NileRuntimeEnvironment:
             address_or_alias = normalize_number(address_or_alias)
         return next(deployments.load_class(address_or_alias, self.network))
 
-    def get_or_deploy_account(self, signer, watch_mode=None):
+    def get_or_deploy_account(self, signer):
         """Get or deploy an Account contract."""
-        return Account(signer=signer, network=self.network, watch_mode=watch_mode)
+        return Account(signer=signer, network=self.network)
 
     def get_accounts(self, predeployed=False):
         """Retrieve and manage deployed accounts."""


andrew-fleming (approved):
Looks good to me! We should create another issue for adding watch_mode to the NRE

ericglau (approved):

martriay:
I don't think this patches the issue, since the script in the original issue uses `Account` directly. although we could remove it from there too, I think it's better to properly handle the error. we've been experiencing some instability in the codebase and I think patches/hotfixes like this will only make this worse.

i suggest raising an exception *after* unregistration on tx failure, and let users handle that exception from the script. what do you think?

ericnordelo:
> I don't think this patches the issue, since the script in the original issue uses Account directly. although we could remove it from there too, I think it's better to properly handle the error. we've been experiencing some instability in the codebase and I think patches/hotfixes like this will only make this worse.

The script tested with `watch_mode` first, and the second issue is a combination of `watch_mode` and the not accomplished invariant. For me, the script without `watch_mode` is working, not sure if I missed something but I think it patches the issue. Your proposed solution seems good, but we need to refactor the integration in more commands, and think about the design. I don't think this is a tiny update, and I'm already working on a refactor to better handle transactions (important for adding queries to declare, and deploy_account). I can work on this, but I rather wait till I finish what I'm working on if you agree (switching context is a costly task).

martriay:
this doesn't path the issue because users can still pass `watch_mode` to `Account`, it's in a script but not an nre function. that's what Eric does in his script.

ericnordelo:
> this doesn't patch the issue because users can still pass `watch_mode` to `Account`, it's available in scripts but not part of NRE's interface, which you patched. that's what Eric does in his script.

Ohh, true, I didn't think it through. 

> agree on solid design, we're in strong debt. but picking between hotfixes, I lean towards a not-thoroughly-designed exception that removing the feature for good -- waiting for txs is very useful when scripting.

Ok, I will check this issue first before continuing with what I'm working on.

ericnordelo:
> i suggest raising an exception _after_ unregistration on tx failure, and let users handle that exception from the script. what do you think? some better version of:

I was checking the code trying to implement a fairly quick fix from this suggestion, but we can't just raise an exception from the places we currently check status because this flow is also used for CLI (I would need to implement a pattern for exception handling in CLI to allow this, which is not a small fix, and is also not the best design to raise an exception from NRE IMO).

I can't think of a good approach to fix this that is small enough with the current design :(.

> although we could remove it from there too, I think it's better to properly handle the error. we've been experiencing some instability in the codebase and I think patches/hotfixes like this will only make this worse.

I agree that hot fixes like this will make this worse, but I don't think we can properly handle the issue without following a proper design (not a small fix). IMO, any quick fix we can come up with is going to be one of these patches/hotfixes you mention. I'd suggest handling this properly even if is not part of this fix-release (and for now either remove the watch_mode from NRE, or ignore it until is properly fixed). 

Any suggestion on how to handle this properly in a quick fix is welcome of course.

Issue #278, opened 2022-11-07
Separate Account initialization from Account deployment.


ericnordelo:
Currently, we are deploying the Account if some conditions are met directly from initialization. This has some problems like nile automatically trying to deploy an Account in the send command if the account doesn't exist (not sure if this was the expected behavior), and also, deployment attempts when Account is used in NRE. I think it would be helpful for clarity and development to decouple these processes.

andrew-fleming:
Also, tackling this issue should include adding `deploy_account` to Nile's documentation of the Account API (following the merge of #334)

PR #315, opened 2022-11-29
Add `nile-account` flag to set Nile artifacts path

### LINKED ISSUE #309: No such file or directory: 'artifacts/Account.json' when querying transaction
From a new project generated by `nile init`, run `nile node` then `nile declare PRIVATE_KEY contract`. This fails with the below which is the same as step 6 of https://github.com/OpenZeppelin/nile/issues/308.
```
🚀 Deploying Account
⏳ ️Deployment of Account successfully sent at 0x02e2adf85e7439821a4f15e61558a116ac89a82f28dcf98fc9d74aab740c59ec
🧾 Transaction hash: 0x6e12304cf4b97cfc0530637a4c563180a84d4a109307574967490013205c4be
📦 Registering 0x02e2adf85e7439821a4f15e61558a116ac89a82f28dcf98fc9d74aab740c59ec in localhost.deployments.txt
  File "/Users/eric/git/nileproj3/env/bin/nile", line 8, in <module>
    sys.exit(cli())
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1157, in __call__
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/anyio/_core/_eventloop.py", line 70, in run
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 292, in run
  File "/opt/homebrew/Cellar/python@3.9/3.9.15/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/runners.py", line 44, in run
  File "/opt/homebrew/Cellar/python@3.9/3.9.15/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 287, in wrapper
    return await func(*args)
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1160, in _main
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1076, in main
    rv = await self.invoke(ctx)
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1687, in invoke
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1434, in invoke
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 780, in invoke
    rv = await rv
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/nile/cli.py", line 176, in declare
    account = await Account(signer, network)
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/nile/core/account.py", line 40, in __new__
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/nile/core/account.py", line 104, in __init__
AssertionError
```

Then run `nile declare PRIVATE_KEY contract` again. This causes an error with `No such file or directory: 'artifacts/Account.json'` which occurs when the command is querying the transaction.  (This also occurs when you run `nile status 0x072757b6b4922caeb0d478481635494d88d4be03386f3cffbe1858f00974145`)
```
🚀 Declaring contract
⏳ Successfully sent declaration of contract as 0x072757b6b4922caeb0d478481635494d88d4be03386f3cffbe1858f00974145
🧾 Transaction hash: 0x2f9325203a92187e342565810d368c164a2790e9e51c73e469bc6a8625a28cf
📦 Registering 0x072757b6b4922caeb0d478481635494d88d4be03386f3cffbe1858f00974145 in localhost.declarations.txt
⏳ Querying the network for transaction status...
❌ Transaction status: REJECTED
🧾 Found contracts: ['0x02e2adf85e7439821a4f15e61558a116ac89a82f28dcf98fc9d74aab740c59ec:artifacts/Account.json']
⏳ Querying the network with identified contracts...
  File "/Users/eric/git/nileproj3/env/bin/nile", line 8, in <module>
    sys.exit(cli())
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1157, in __call__
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/anyio/_core/_eventloop.py", line 70, in run
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 292, in run
  File "/opt/homebrew/Cellar/python@3.9/3.9.15/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/runners.py", line 44, in run
  File "/opt/homebrew/Cellar/python@3.9/3.9.15/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 287, in wrapper
    return await func(*args)
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1160, in _main
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1076, in main
    rv = await self.invoke(ctx)
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1687, in invoke
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 1434, in invoke
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/asyncclick/core.py", line 780, in invoke
    rv = await rv
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/nile/cli.py", line 177, in declare
    await account.declare(
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/nile/core/account.py", line 170, in declare
    return await declare(
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/nile/core/declare.py", line 50, in declare
    tx_status = await status(tx_hash, network, watch_mode)
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/nile/utils/status.py", line 43, in status
    error_message = await debug_message(
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/nile/utils/debug.py", line 42, in debug_message
    return await execute_call(
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/nile/starknet_cli.py", line 30, in execute_call
    return await capture_stdout(cmd(args=args, command_args=command_args))
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/nile/starknet_cli.py", line 47, in capture_stdout
    await func
  File "/Users/eric/git/nileproj3/env/lib/python3.9/site-packages/starkware/starknet/cli/starknet_cli.py", line 441, in tx_status
    contracts[addr] = Program.load(data=json.load(open(path.strip()))["program"])
FileNotFoundError: [Errno 2] No such file or directory: 'artifacts/Account.json'
```

andrew-fleming:
This PR proposes to add the `nile-account` flag to `account.declare` which will create an overriding_path to Nile's precompiled artifacts. This will remove the need to manually set the path for Account declarations.

This PR just needs tests.

Resolves #309.

### DIFF
--- a/src/nile/cli.py
+++ b/src/nile/cli.py
@@ -159,6 +159,7 @@ async def deploy(
 @click.option("--max_fee", nargs=1)
 @click.option("--alias")
 @click.option("--overriding_path")
+@click.option("--nile_account", is_flag=True)
 @network_option
 @mainnet_token_option
 @watch_option
@@ -171,6 +172,7 @@ async def declare(
     alias,
     overriding_path,
     token,
+    nile_account,
 ):
     """Declare a StarkNet smart contract through an Account."""
     account = await Account(signer, network)
@@ -181,6 +183,7 @@ async def declare(
         overriding_path=overriding_path,
         mainnet_token=token,
         watch_mode=watch_mode,
+        nile_account=nile_account,
     )
 
 
--- a/src/nile/core/account.py
+++ b/src/nile/core/account.py
@@ -106,8 +106,7 @@ class Account(AsyncObject):
     async def deploy(self, salt=None, max_fee=None, query_type=None, watch_mode=None):
         """Deploy an Account contract for the given private key."""
         index = accounts.current_index(self.network)
-        pt = os.path.dirname(os.path.realpath(__file__)).replace("/core", "")
-        overriding_path = (f"{pt}/artifacts", f"{pt}/artifacts/abis")
+        overriding_path = set_nile_artifacts_path()
 
         class_hash = get_account_class_hash("Account")
         salt = 0 if salt is None else normalize_number(salt)
@@ -152,9 +151,14 @@ class Account(AsyncObject):
         overriding_path=None,
         mainnet_token=None,
         watch_mode=None,
+        nile_account=False,
     ):
         """Declare a contract through an Account."""
         max_fee, nonce, _ = await self._process_arguments(max_fee, nonce)
+
+        if nile_account:
+            assert overriding_path is None
+            overriding_path = set_nile_artifacts_path()
 
         contract_class = get_contract_class(
             contract_name=contract_name, overriding_path=overriding_path
@@ -174,6 +178,7 @@ class Account(AsyncObject):
             alias=alias,
             network=self.network,
             max_fee=max_fee,
+            overriding_path=overriding_path,
             mainnet_token=mainnet_token,
             watch_mode=watch_mode,
         )
@@ -187,6 +192,7 @@ class Account(AsyncObject):
         alias,
         max_fee=None,
         deployer_address=None,
+        overriding_path=None,
         abi=None,
         watch_mode=None,
     ):
@@ -204,6 +210,7 @@ class Account(AsyncObject):
             alias,
             deployer_address,
             max_fee,
+            overriding_path=overriding_path,
             abi=abi,
             watch_mode=watch_mode,
         )
@@ -294,3 +301,10 @@ def get_counterfactual_address(salt=None, calldata=None, contract="Account"):
         constructor_calldata=calldata,
         deployer_address=0,
     )
+
+
+def set_nile_artifacts_path():
+    """Set path to find Nile precompiled artifacts."""
+    pt = os.path.dirname(os.path.realpath(__file__)).replace("/core", "")
+    overriding_path = (f"{pt}/artifacts", f"{pt}/artifacts/abis")
+    return overriding_path
--- a/src/nile/core/account.py
+++ b/src/nile/core/account.py
@@ -304,7 +304,7 @@ def get_counterfactual_address(salt=None, calldata=None, contract="Account"):
 
 
 def set_nile_artifacts_path():
-    """Set path to find Nile precompiled artifacts."""
+    """Set path to find Nile's precompiled artifacts."""
     pt = os.path.dirname(os.path.realpath(__file__)).replace("/core", "")
     overriding_path = (f"{pt}/artifacts", f"{pt}/artifacts/abis")
     return overriding_path
--- a/tests/commands/test_account.py
+++ b/tests/commands/test_account.py
@@ -23,6 +23,7 @@ MAX_FEE = 10
 SALT = 444
 SIGNATURE = [111, 222]
 CLASS_HASH = 12345
+PATH = ("src/nile/artifacts", "src/nile/artifacts/abis")
 
 
 @pytest.fixture(autouse=True)
@@ -140,6 +141,7 @@ async def test_declare(mock_declare, mock_get_class, mock_hash, mock_deploy):
         network=NETWORK,
         alias=alias,
         max_fee=max_fee,
+        overriding_path=overriding_path,
         mainnet_token=None,
         watch_mode=None,
     )
@@ -149,11 +151,18 @@ async def test_declare(mock_declare, mock_get_class, mock_hash, mock_deploy):
 @pytest.mark.parametrize("deployer_address", [None, 0xDE0])
 @pytest.mark.parametrize("watch_mode", [None, "debug"])
 @pytest.mark.parametrize("abi", [None, "TEST_ABI"])
+@pytest.mark.parametrize("overriding_path", [None, PATH])
 @patch("nile.core.account.deploy_account", return_value=(MOCK_ADDRESS, MOCK_INDEX))
 @patch("nile.core.deploy.get_class_hash", return_value=0x434343)
 @patch("nile.core.account.deploy_with_deployer")
 async def test_deploy_contract(
-    mock_deploy_contract, mock_get_class, mock_deploy, abi, watch_mode, deployer_address
+    mock_deploy_contract,
+    mock_get_class,
+    mock_deploy,
+    overriding_path,
+    abi,
+    watch_mode,
+    deployer_address,
 ):
     account = await Account(KEY, NETWORK)
 
@@ -172,8 +181,9 @@ async def test_deploy_contract(
         alias,
         max_fee,
         deployer_address,
-        abi,
-        watch_mode,
+        overriding_path=overriding_path,
+        abi=abi,
+        watch_mode=watch_mode,
     )
 
     if deployer_address is None:
@@ -189,6 +199,7 @@ async def test_deploy_contract(
         alias,
         deployer_address,
         max_fee,
+        overriding_path=overriding_path,
         abi=abi,
         watch_mode=watch_mode,
     )
--- a/src/nile/core/deploy.py
+++ b/src/nile/core/deploy.py
@@ -101,7 +101,7 @@ async def deploy_contract(
         _salt = compute_hash_chain(data=[account.address, salt])
         deployer_for_address_generation = deployer_address
 
-    class_hash = get_class_hash(contract_name=contract_name)
+    class_hash = get_class_hash(contract_name, overriding_path)
 
     address = calculate_contract_address_from_hash(
         _salt, class_hash, calldata, deployer_for_address_generation
--- a/tests/commands/test_deploy.py
+++ b/tests/commands/test_deploy.py
@@ -19,16 +19,6 @@ def tmp_working_dir(monkeypatch, tmp_path):
     return tmp_path
 
 
-EXP_SALTS = [
-    2575391846029882800677169842619299590487820636126802982795520479739126412818,
-    2557841322555501036413859939246042028937187876697248667793106475357514195630,
-]
-EXP_CLASS_HASHES = [
-    0x434343,
-    0x464646,
-    0x494949,
-    0x525252,
-]
 MOCK_ACC_ADDRESS = 0x123
 MOCK_ACC_INDEX = 0
 CONTRACT = "contract"
@@ -46,6 +36,24 @@ CALL_OUTPUT = [ADDRESS, TX_HASH]
 SIGNATURE = [111, 333]
 SALT = 555
 FEE = 666
+
+DEPLOY_ARGS = [
+    # name, salt, unique, calldata, alias, deployer, max_fee, [overriding_path], [abi]
+    [CONTRACT, 0, True, [], ALIAS, 0x424242, 5],
+    [CONTRACT, 1, False, [1], ALIAS, 0x454545, 0],
+    [CONTRACT, 3, True, [1, 2], ALIAS, 0x484848, 0],
+    [CONTRACT, 3, False, [1, 2, 3], ALIAS, 0x515151, 0, None, ABI_OVERRIDE],
+]
+EXP_SALTS = [
+    2575391846029882800677169842619299590487820636126802982795520479739126412818,
+    2557841322555501036413859939246042028937187876697248667793106475357514195630,
+]
+EXP_CLASS_HASHES = [
+    0x434343,
+    0x464646,
+    0x494949,
+    0x525252,
+]
 
 
 @pytest.mark.asyncio
@@ -125,35 +133,25 @@ async def test_deploy(mock_register, mock_parse, caplog, args, cmd_args, exp_abi
     "args, exp_class_hash, exp_salt, exp_abi",
     [
         (
-            [CONTRACT, 0, True, [], ALIAS, 0x424242, 5],  # args
+            DEPLOY_ARGS[0],
             EXP_CLASS_HASHES[0],
             EXP_SALTS[0],
             ABI,
         ),
         (
-            [CONTRACT, 1, False, [1], ALIAS, 0x454545, 0],  # args
+            DEPLOY_ARGS[1],
             EXP_CLASS_HASHES[1],
             1,
             ABI,
         ),
         (
-            [CONTRACT, 3, True, [1, 2], ALIAS, 0x484848, 0],  # args
+            DEPLOY_ARGS[2],
             EXP_CLASS_HASHES[2],
             EXP_SALTS[1],
             ABI,
         ),
         (
-            [
-                CONTRACT,
-                3,
-                False,
-                [1, 2, 3],
-                ALIAS,
-                0x515151,
-                0,
-                None,
-                ABI_OVERRIDE,
-            ],  # args
+            DEPLOY_ARGS[3],
             EXP_CLASS_HASHES[3],
             3,
             ABI_OVERRIDE,
@@ -203,6 +201,7 @@ async def test_deploy_contract(
             max_fee=max_fee,
         )
         mock_register.assert_called_once_with(exp_address, exp_abi, NETWORK, ALIAS)
+        mock_return_account.assert_called_once_with(CONTRACT, None)
 
         # check logs
         assert f"🚀 Deploying {CONTRACT}" in caplog.text


ericnordelo:
Hi Andrew! This approach to declaring looks great!

> This PR just needs tests.
> Resolves https://github.com/OpenZeppelin/nile/issues/309.

I just want to mention that 309 is not completely solved by it. We still have the problem with status, which tries to "infer" the artifact of the contract instead of getting it from the **deployments.txt** file (using `execute_call` in new **starknet_cli.py** logic):

https://github.com/OpenZeppelin/nile/blob/98ce378842eb5d404f82e3bc2d082dd06091b665/src/nile/starknet_cli.py#L76-L85

Every time a transaction is sent through an account (most of the time) status is going to fail to try to get the Account artifact, which is already registered correctly in **deployments.txt**.

Keeping in mind that status uses the contracts registered in this file, and this file contains the artifacts, I think it makes sense to use the artifacts already registered instead of trying to pass `overrwriting_path` down the flow.

martriay:
Agree on reading the deployments file. This file exists for these reasons, it also makes custom account integration easier down the road.

> Why are we registering the artifact in deployments.txt, if we are going to assume they are always inside the nile artifact folder

We should use the artifact, but a possible reason to register it if not would be to maintain the file structure/invariants.

andrew-fleming:
> We still have the problem with status, which tries to "infer" the artifact of the contract instead of getting it from the deployments.txt file (using execute_call in new starknet_cli.py logic)

This is partially correct^. The issue is actually with `Debug` in that `Debug` iterates over the `.cairo` contracts in order to highlight where the error comes from. The `FileNotFoundError: [Errno 2] No such file or directory: 'artifacts/Account.json'` is coming from `_get_contracts_data` with the code below for reference:

```
def _get_contracts_data(contracts_file, network, addresses):
    file = contracts_file or f"{network}.{DEPLOYMENTS_FILENAME}"
    # contracts_file should already link to compiled contracts and not ABIs
    to_contract = (lambda x: x) if contracts_file else _abi_to_build_path
    contracts = _locate_error_lines_with_abis(file, addresses, to_contract)
    return contracts
```

Being that we only have the artifacts for `Account`, this is an issue with the proposed solution. I tested this even further by installing OZ Contracts, compiling `Account`, and adding the `Account.cairo` path and it works as I thought.

----

> Every time a transaction is sent through an account (most of the time) status is going to fail to try to get the Account artifact, which is already registered correctly in deployments.txt.

This is not entirely accurate. While it won't be able to iterate over the precompiled contracts, I wouldn't say that the status is going to fail. There's already handling for contracts not found https://github.com/OpenZeppelin/nile/blob/main/src/nile/utils/debug.py#L30-L37

```
    contracts = _get_contracts_data(contracts_file, network, addresses)

    if not contracts:
        logging.warning(
            "🛑 The transaction was rejected but no contract data is locally "
            "available to improve the error message."
        )
        return error_message
```

which looks like this:

```
🚀 Deploying contract
Invoke transaction was sent.
Contract address: 0x04d3f77f305c02d158f159a91e00f4562e8697b7025559aa5f0497446c3bd5de
Transaction hash: 0x7e98b1e34107b963dfced6f57f86b005007bde795423b5ae3d42455a60c48c1
⏳ ️Deployment of contract successfully sent at 0x00dcfac954d8ac11b0de8107520b63c0f1a1bb5928cdec050878370327ce41a4
🧾 Transaction hash: 0x7e98b1e34107b963dfced6f57f86b005007bde795423b5ae3d42455a60c48c1
📦 Registering deployment as my_contract in localhost.deployments.txt
⏳ Querying the network for transaction status...
❌ Transaction status: REJECTED
🛑 The transaction was rejected but no contract data is locally available to improve the error message.
🧾 Error message:
/home/cairo-contracts/env/lib/python3.8/site-packages/starkware/starknet/common/syscalls.cairo:52:5: Error at pc=0:81:
Got an exception while executing a hint.
Cairo traceback (most recent call last):
src/openzeppelin/account/presets/Account.cairo:128:6: (pc=0:731)
src/openzeppelin/account/presets/Account.cairo:143:36: (pc=0:677)
/home/cairo-contracts/src/openzeppelin/account/library.cairo:201:30: (pc=0:291)
/home/cairo-contracts/src/openzeppelin/account/library.cairo:218:19: (pc=0:314)

Error in the called contract (0x41a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf):
Error at pc=0:32:
Got an exception while executing a hint.
Cairo traceback (most recent call last):
Unknown location (pc=0:174)
Unknown location (pc=0:127)

  ...
    raise get_stark_exception_on_undeclared_contract(class_hash=class_hash)
starkware.starkware_utils.error_handling.StarkException: (500, {'code': <StarknetErrorCode.UNDECLARED_CLASS: 49>, 'message': 'Class with hash 0x72757b6b4922caeb0d478481635494d88d4be03386f3cffbe1858f00974145 is not declared.'})
```

----

I will include a solution for better error handling in cases where the user tries to declare a contract through an undeclared/undeployed sender account. What is not handled, however, is the unregistration of the account through this process (the declaration is nevertheless removed successfully). I'm still looking into the reason.

IMO this whole process highlights why we should prioritize separating account initialization and deployment (#278). With this separation, I believe we'll have stronger control over the account flow and error handling thus better UX.

ericnordelo:
Thanks for the context @andrew-fleming!

> The issue is actually with `Debug` in that `Debug` iterates over the `.cairo` contracts in order to highlight where the error comes from

Yes! My bad. The issue is with debug, not with status, I didn't explain myself well. I said status because debug is a type of status, and is the default that runs (issue only happens if the underlying transaction reverts).

> The `FileNotFoundError: [Errno 2] No such file or directory: 'artifacts/Account.json'` is coming from `_get_contracts_data` with the code below for reference:

Not sure what we are doing different, but following the same steps provided by Eric in the issue, the traceback of the error I received is this one:

 ```
 File "/Users/ericnordelo/Documents/Projects/cairo/test_collection/env/lib/python3.9/site-packages/nile/utils/status.py", line 43, in status
    error_message = await debug_message(
  File "/Users/ericnordelo/Documents/Projects/cairo/test_collection/env/lib/python3.9/site-packages/nile/utils/debug.py", line 42, in debug_message
    return await execute_call(
  File "/Users/ericnordelo/Documents/Projects/cairo/test_collection/env/lib/python3.9/site-packages/nile/starknet_cli.py", line 30, in execute_call
    return await capture_stdout(cmd(args=args, command_args=command_args))
  File "/Users/ericnordelo/Documents/Projects/cairo/test_collection/env/lib/python3.9/site-packages/nile/starknet_cli.py", line 47, in capture_stdout
    await func
  File "/Users/ericnordelo/Documents/Projects/cairo/test_collection/env/lib/python3.9/site-packages/starkware/starknet/cli/starknet_cli.py", line 441, in tx_status
    contracts[addr] = Program.load(data=json.load(open(path.strip()))["program"])
FileNotFoundError: [Errno 2] No such file or directory: 'artifacts/Account.json'
```

Here you can see that the `FileNotFoundError` error is not coming from `_get_contracts_data`, but from the `execute_call` call in line 42. That's why I assumed the issue was coming from the `straknet_cli.py` design. I'm testing in main branch, maybe this i fixed already in this PR?

> This is not entirely accurate. While it won't be able to iterate over the precompiled contracts, I wouldn't say that the status is going to fail. There's already handling for contracts not found

The status is failing already for declare in the issue right? what is the difference with other commands going through accounts? I think is important to discover why your traceback and mine are different in the first place for me to understand this point.

**Comment by martriay** [comment]
i get the same as Eric. so in short, if i understood correctly:

* nile custom artifacts are not properly picked up by artifact-based commands, like account's deploy/declare/send. i guess this affects not just account but also ERC20 (eth token) functions?
* Andrew accurately raises that once the artifact is in place, debug would still not work since the account’s source code is not available for processing

i assume Andrew's getting a different error either because he added/compiled a new artifact, or because he has a local fix already

**Comment by andrew-fleming** [comment]
@ericnordelo 
> The issue is with debug, not with status, I didn't explain myself well. I said status because debug is a type of status, and is the default that runs (issue only happens if the underlying transaction reverts).

Haha no worries! This was a tricky issue, and I just wanted to be as precise as possible.

> Here you can see that the FileNotFoundError error is not coming from _get_contracts_data, but from the execute_call call in line 42. That's why I assumed the issue was coming from the straknet_cli.py design.

So the assumption is correct in that the error is raised by `execute_call`. I mentioned that the error is coming from `_get_contracts_data` because that's where the problem is stemming from (it's just not raised until `execute_call`). That's why it was tricky because I also assumed we could somehow set the path to the artifacts for instances where they're returning errors.

Also @martriay ^

**Comment by martriay** [comment]
But the error says the missing file is `Account.json`, not the Cairo one. As you say that's properly handled, so what's the error stemming there? and how does that explain the artifact error we get?

**Comment by andrew-fleming** [comment]
> But the error says the missing file is Account.json, not the Cairo one. As you say that's properly handled, so what's the error stemming there? and how does that explain the artifact error we get?

Excellent question. The Account.json is being set in `_abi_to_build_path`...and setting the path there directly solves that error!

File: src/nile/cli.py [hunk 1]
```diff
@@ -171,16 +172,18 @@ async def declare(
...
-    account = await Account(signer, network)
+    account = await Account(signer, network, watch_mode=watch_mode)
     await account.declare(
         contract_name,
         alias=alias,
         max_fee=max_fee,
         overriding_path=overriding_path,
         mainnet_token=token,
         watch_mode=watch_mode,
+        nile_account=nile_account,
```
martriay, line 186: what about making it a constructor argument and object attribute? i.e. `self.is_nile_account`

it could be leaner if we eventually need it for other operations, and it would debloat the function's arguments. although right now it's the same since it's just here

src/nile/core/account.py
```diff
@@ -152,10 +151,15 @@ async def declare(
         overriding_path=None,
         mainnet_token=None,
         watch_mode=None,
+        nile_account=False,
     ):
         """Declare a contract through an Account."""
         max_fee, nonce, _ = await self._process_arguments(max_fee, nonce)
 
+        if nile_account:
+            assert overriding_path is None
```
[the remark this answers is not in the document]
andrew-fleming: If it's a nile_account, an overriding_path wouldn't make sense because the path to the artifacts lies within the package. The assertion isn't absolutely necessary, but the user might get the wrong idea about where the account is coming from...an error message with the assertion would be helpful

src/nile/utils/debug.py
```diff
@@ -39,23 +39,42 @@ async def debug_message(error_message, tx_hash, network, contracts_file=None):
...
+        "hash": hex_class_hash(tx_hash),
+        "contracts": ",".join(contracts),
+        "error_message": True,
+    }
+
+    try:
+        return await execute_call(**execute_args)
+    except FileNotFoundError:
+        # Change path in contracts to locate Nile artifacts
+        contracts = _get_contracts_data(contracts_file, network, addresses, True)
```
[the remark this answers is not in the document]
andrew-fleming: Yeah, that should work. I was initially worried about conflicting "Account"s (with how common the contract name is), but that shouldn't be an issue now with the `nile_account`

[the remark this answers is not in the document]
martriay: we could also have a different name for Nile's

src/nile/cli.py
andrew-fleming: We can maybe add this in the future, but I don't think this would work here. The point of `nile_account` is to direct the path to the Nile artifacts. Really, it's just for declaring a Nile account, so we need a way to adjust the path for this specific type of declaration. In other words, when an account wants to declare a Nile account, use the flag to change the path. For other declarations, Nile assumes the local `artifacts/` path.

IMO a better fix would be to iterate both paths checking for the target contract. That way, we don't have to worry about any of this. I already have a working example of this improvement, but for the sake of getting this shipped, I decided not to include it

src/nile/core/account.py
```diff
@@ -297,3 +301,10 @@ def get_counterfactual_address(salt=None, calldata=None, contract="Account"):
         constructor_calldata=calldata,
         deployer_address=0,
     )
+
+
+def set_nile_artifacts_path():
```
ericnordelo, line 306:
```suggestion
def get_nile_artifacts_path():
```
A getter with `set` in the name is a little confusing.

src/nile/core/account.py
```diff
@@ -297,3 +301,10 @@ def get_counterfactual_address(salt=None, calldata=None, contract="Account"):
...
+    """Set path to find Nile's precompiled artifacts."""
+    pt = os.path.dirname(os.path.realpath(__file__)).replace("/core", "")
```
ericnordelo, line 308: Any reason to not import `pt` from `nile.common`?

ericnordelo (commented):
Looking good @andrew-fleming ! Left a couple of comments. A refactor for the `overriding_path` handling flow would be helpful I think, but that is out of the scope of this PR.

src/nile/core/account.py
ericnordelo: If we use it, I would rename `pt` to `nile_root_path` or something similar.

src/nile/core/account.py
andrew-fleming: Haha yeah, I went back and forth on this.

src/nile/core/account.py
andrew-fleming: > Any reason to not import pt from nile.common?

Nope, good call!

src/nile/core/account.py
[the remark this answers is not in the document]
martriay: but wouldn't this let users deploy custom accounts?

tests/commands/test_account.py
```diff
@@ -7,11 +7,13 @@
...
     ABIS_DIRECTORY,
     BUILD_DIRECTORY,
+    NILE_ABIS_DIR,
+    NILE_BUILD_DIR,
     QUERY_VERSION,
     TRANSACTION_VERSION,
     UNIVERSAL_DEPLOYER_ADDRESS,
 )
-from nile.core.account import Account, set_nile_artifacts_path
+from nile.core.account import Account, get_nile_artifacts_path
```
martriay, line 16: this seems like a `common` function, or maybe even just a constant

tests/test_accounts.py
```diff
@@ -0,0 +1,73 @@
+"""Tests for accounts file."""
```
martriay, line 1: then i'd call it `tests/test_accounts_file.py`

martriay (approved):
looks good! in the future we should add more tests for account file management, like `register`

src/nile/core/account.py
```diff
@@ -297,3 +303,8 @@ def get_counterfactual_address(salt=None, calldata=None, contract="Account"):
         constructor_calldata=calldata,
         deployer_address=0,
     )
+
+
+def get_nile_artifacts_path():
```
ericnordelo, line 308: This could be a constant.

ericnordelo (dismissed):
Looking good to go! Left a small suggestion.

ericnordelo (approved):
LGTM!

PR #318, opened 2022-11-29
Allow hex params in calldata for deployments through Account

### LINKED ISSUE #317: nile deploy command doesn't accept hex arguments in calldata
The `Account.deploy_contract()` method is not processing calldata (normalizing arguments) like send. Therefore calldata with hex values result in a python exception.

ericnordelo:
Fixes #317

### DIFF
--- a/src/nile/core/account.py
+++ b/src/nile/core/account.py
@@ -195,6 +195,8 @@ class Account(AsyncObject):
             deployer_address or UNIVERSAL_DEPLOYER_ADDRESS
         )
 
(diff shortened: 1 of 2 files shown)


martriay (approved):

martriay (changes requested):
can you add tests for it? the ones that would have prevented us from releasing without this :)

martriay (approved):

andrew-fleming (approved):
Looks good to go!

PR #327, opened 2022-12-12
Update goerli2 chain id

### LINKED ISSUE #321: update goerli2 chain id
the chain id of goerli2 is now `SN_GOERLI2`

ericnordelo:
Fixes #321


src/nile/signer.py
```diff
@@ -20,11 +20,14 @@ def __init__(self, private_key, network="testnet"):
...
-            else StarknetChainId.TESTNET.value
-        )
+        if network == "mainnet":
+            self.chain_id = StarknetChainId.MAINNET.value
+        else:
+            self.chain_id = (
+                StarknetChainId.TESTNET2.value
+                if network == "goerli2"
+                else StarknetChainId.TESTNET.value
+            )
```
martriay, line 30:
```suggestion
        self.chain_id = {
            "mainnet": StarknetChainId.MAINNET.value,
            "goerli2": StarknetChainId.TESTNET2.value,
            "goerli": StarknetChainId.TESTNET.value
        }[network or "goerli"]
```

what do you think of something like this?

tests/test_signer.py
```diff
@@ -108,14 +116,14 @@ async def test_execute():
...
+@pytest.mark.parametrize("network", NETWORKS)
 @pytest.mark.asyncio
-async def test_chain_id():
-    mainnet = Signer(PRIVATE_KEY, "mainnet")
-    assert mainnet.chain_id == StarknetChainId.MAINNET.value
-
-    testnet = Signer(PRIVATE_KEY, "testnet")
-    assert testnet.chain_id == StarknetChainId.TESTNET.value
+async def test_chain_id(network):
+    # check all networks
```
martriay, line 122: i think it's confusing to say "all networks" and then check only one, i think the "all" part is implied in the fixture rather than here

tests/test_signer.py
```diff
@@ -108,14 +116,14 @@ async def test_execute():
...
+    signer = Signer(PRIVATE_KEY, network)
+    assert signer.chain_id == NETWORK_CHAIN_ID[network]
 
+    # check with no network
```
martriay, line 126:
```suggestion
```
why is this comment is clarifying?

src/nile/signer.py
ericnordelo: That looks neat.

ericnordelo: We have the issue of not allowing custom network names though. But I think we can fix this when we add the nile.ini file with a network config. For example, I could want to configure a mainnet fork as "localhost-mainnet", and this won't work with this approach unless I add a try/catch block handling custom names.

martriay (approved):

PR #332, opened 2023-01-05
Auto calculate max_fee

### LINKED ISSUE #284: auto calculate max_fee
instead of `0` as the default `max_fee`, nile should estimate the required fee for the users, which today have to guess it.

andrew-fleming:
Resolves #284.

This PR proposes to automatically estimate the max_fee when the max_fee isn't explicitly passed (which would be interpreted as `0`). The fee estimation logic lies in `estimate_fee_when_zero`.

Some notes:

CLI
  - `estimate_fee_when_zero` precedes `tx.execute` in the `run_transaction` function.

NRE
- The NRE includes a new method `execute_tx` which estimates the fee (if zero) and executes the transaction. If accepted, the proposed changes will then offer users two ways to script account txs:

```
    await nre.execute_tx(
        account.send(ETH_TOKEN_ADDRESS, "transfer", [recipient, *amount])
    )

    OR:

    tx = await account.send(ETH_TOKEN_ADDRESS, "transfer", [recipient, *amount])
    max_fee = await tx.estimate_fee()
    await tx.update_fee(max_fee).execute()
```

### DIFF
--- a/src/nile/core/types/account.py
+++ b/src/nile/core/types/account.py
@@ -271,7 +271,16 @@ def _get_signer_and_alias(signer, predeployed_info):
 async def _set_estimated_fee_if_none(max_fee, tx):
     """Estimate max_fee for transaction if max_fee is None."""
     if max_fee is None:
+        logger = logging.getLogger()
+        current_level = logger.level
+
+        # Avoid logging the fee estimation in CLI
+        logger.setLevel(logging.WARNING)
+
         estimated_fee = await tx.estimate_fee()
+
+        logger.setLevel(current_level)
+
         tx.update_fee(estimated_fee)
 
 


src/nile/nre.py
```diff
@@ -33,6 +33,12 @@ def call(self, address_or_alias, method, params=None, abi=None):
             address_or_alias, "call", method, params, self.network, abi=abi
         )
 
+    async def execute_tx(self, tx_wrapper, watch_mode=None):
```
ericnordelo, line 36: Keeping in mind that this method is modifying the tx (not only executing it), I would rename it to something more descriptive. Not sure what name to use though, maybe `set_estimated_fee_and_execute`? This name doesn't take into account that if the fee is different than 0 the estimated one won't be used, so I think is still confusing.

Keeping in mind that execute_tx updates `tx.hash`, `tx.max_fee`, and `tx.query_hash`, users can get unexpected behavior on the script by calling this `execute_tx` method without knowing that it modifies the tx object.

I'm more in favor of keeping only this way, which is more explicit and easier to follow because of that:
```
tx = await account.send(ETH_TOKEN_ADDRESS, "transfer", [recipient, *amount])
max_fee = await tx.estimate_fee()
await tx.update_fee(max_fee).execute()
```

ericnordelo (commented):
Looking good Andrew! Glad to see no big changes were required from 320. Left some comments.

src/nile/nre.py
```diff
@@ -33,6 +33,12 @@ def call(self, address_or_alias, method, params=None, abi=None):
...
+        """Add estimated max_fee (if max_fee is zero) and execute transaction."""
+        tx = await tx_wrapper
```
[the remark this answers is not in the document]
andrew-fleming: Yeah, that sounds good to me. I was trying just trying to simplify as much as possible, I agree that your proposed solution looks cleaner

src/nile/nre.py
andrew-fleming: > Keeping in mind that this method is modifying the tx (not only executing it), I would rename it to something more descriptive

Agreed. The idea was to have a simple and easy-to-use method. It's probably better to have `set_estimated_fee_and_execute` if we keep this approach

> users can get unexpected behavior on the script by calling this execute_tx method without knowing that it modifies the tx object

> I'm more in favor of keeping only this way, which is more explicit and easier to follow because of that:

My only contention with this consists of the "auto calculate fee" issue itself because we would be asking the user to set up the `update_fee` and it wouldn't exactly be automated in the NRE (though, the flow makes it quite easy). I suppose the question is: do you think the `tx.update_fee` flow is enough to satisfy the issue?

ericnordelo: Got you. I think having the current design, in NRE `update_fee` is enough to satisfy because issues by misleading tx hashes after execution weigh more than the benefit over the two extra lines (in CLI happens transparently).

Maybe we could implement an "auto fee" mechanism in Transaction initialization that would initialize the transaction with the estimated fee, transparently to the user (if the max_fee in construction time is set to -1 or "auto" in initialization for example), what do you think?

andrew-fleming: I think that's a better idea! And it's implemented, I just need to update tests

src/nile/core/types/transactions.py
```diff
@@ -67,8 +67,12 @@ def __post_init__(self):
         # Validate the transaction object
         self._validate()
 
-    async def execute(self, signer, max_fee=None, watch_mode=None, **kwargs):
+    async def execute(self, signer, watch_mode=None, **kwargs):
         """Execute the transaction."""
+        if self.max_fee == 0:
+            max_fee = await self.estimate_fee(signer)
+            self.update_fee(max_fee)
```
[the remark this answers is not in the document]
ericnordelo: I also think that we should let the user set the `max_fee` as 0, that's why I think "auto" would be a better option for auto fee estimation. Then we know the user won't receive a different transaction from the one he was expecting (it could be one with max_fee equal 0, even if it doesn't make much sense to us, it could make sense).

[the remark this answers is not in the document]
andrew-fleming: Ah forgive me, I lost sight of the tx.hash changing. The issue I was running into with that approach is `estimate_fee` is async, so I think executing this in during construction (with the sync `__post_init__`) won't work without getting into hacky territory. Because of the async issue, I don't think we can pass `auto` or `-1` to the `Transaction` during construction either.

Two options that I see are:
1. Have the users use `tx.update_fee` which IMO isn't automating the fee but as you mentioned, maybe it's enough
2. Automate the fee in the account methods which keeps the tx.hash consistent for the user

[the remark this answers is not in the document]
ericnordelo: Number 2 sounds great. Returning the tx_wrapper with the estimated fee if max_fee is left to None in account methods.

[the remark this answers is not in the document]
andrew-fleming: On it!

src/nile/core/types/account.py
```diff
@@ -255,6 +268,13 @@ def _get_signer_and_alias(signer, predeployed_info):
     return signer, alias
 
 
+async def _set_estimated_fee_if_none(max_fee, tx):
+    """Estimate max_fee for transaction if max_fee is None."""
+    if max_fee is None:
+        estimated_fee = await tx.estimate_fee()
+        tx.update_fee(estimated_fee)
```
ericnordelo, line 275: I think with this we are good to go, just a small suggestion for organization.

Keeping in mind we are calling this for all the account methods returning tx wrappers. Can't we add this logic in the `BaseTxWrapper` `__post_init__` method to avoid having to call the method repeatedly in account?

andrew-fleming: That was the idea early on, but we'd need `__post_init__` to be async in order to call the underlying `tx.estimate_fee`. It should be possible to do, but I think it's a bit hacky (a la the Account initialization with the async `self.deploy`)

ericnordelo: Got it, makes sense.

src/nile/core/types/transactions.py
```diff
@@ -47,7 +47,7 @@ class Transaction(ABC):
     """
 
     account_address: int = 0
-    max_fee: int = 0
+    max_fee: Union[int, None] = 0
```
ericnordelo, line 50:
```suggestion
    max_fee: int = 0
```

This will set `max_fee` as 0 even when None is passed (max_fee=None). We don't need to normalize the max_fee in `post_init`.

src/nile/core/types/transactions.py
```diff
@@ -60,14 +60,18 @@ class Transaction(ABC):
 
     def __post_init__(self):
         """Populate pending fields."""
+        # Normalize max_fee if None
+        if self.max_fee is None:
+            self.max_fee = 0
+
```
ericnordelo, line 66:
```suggestion
```

With `max_fee: int = 0` we don't need to normalize the max_fee (will be set to 0 even when None is passed)

ericnordelo (commented):
Almost there! I left a couple of comments to keep the Transaction code as clean as possible, and I think we are good to go.

src/nile/core/types/transactions.py
andrew-fleming: >  If the intention is to support max_fee=max_fee with max_fee equal None from the account methods

That indeed is the intention

> I think is more clear to change the call in the account to max_fee=max_fee or 0 instead.

I agree it's more clear. While this avoids the need to normalize the `max_fee` in `Transaction`, we'd need to create a new variable to save the original `max_fee` passed to the account method (to check if it's None because that's the trigger for the fee estimation).

If you think either way is too much to accommodate the 0 `max_fee` edge case, I can apply the suggestions and keep it simple

ericnordelo: We need an extra variable? I'm thinking in something like this (for each method):

```
        # Create the transaction
        transaction = DeclareTransaction(
            account_address=self.address,
            contract_to_submit=contract_name,
            max_fee=max_fee or 0,
            nonce=nonce,
            network=self.network,
            overriding_path=overriding_path,
        )

        tx_wrapper = DeclareTxWrapper(
            tx=transaction,
            account=self,
            alias=alias,
        )

        await _set_estimated_fee_if_none(max_fee, tx_wrapper)
        return tx_wrapper
```

The change is just `max_fee=max_fee or 0` in the `DeclareTransaction` call.

`_set_estimated_fee_if_none` would still receive None. The transaction doesn't have to, right?

andrew-fleming: Ahhh directly in the function call. Yeah, that's much cleaner. Thank you, sir! Will update

src/nile/core/types/account.py
ericnordelo, line 275:
```suggestion
        logger = logging.getLogger()
        current_level = logger.level

        # Avoid logging the fee estimation in CLI
        logger.setLevel(logging.WARNING)

        estimated_fee = await tx.estimate_fee()

        logger.setLevel(current_level)

        tx.update_fee(estimated_fee)
```

`nile setup ACCOUNT_1 --salt 2 --estimate_fee` is logging the fee estimation twice. And other commands are logging the fee estimation when fee is not set.

ericnordelo (commented):
Almost there :)!

src/nile/core/types/account.py
andrew-fleming: Good catch!

tests/core/types/test_account.py
```diff
@@ -187,9 +188,16 @@ async def test_declare(
...
 async def test_declare_account(
-    mock_tx_wrapper, mock_tx, mock_contract_class, nile_account, overriding_path
+    mock_set_fee,
+    mock_tx_wrapper,
+    mock_tx,
+    mock_contract_class,
+    nile_account,
+    overriding_path,
 ):
+    # mock_tx_wrapper = AsyncMock()
```
martriay, line 200: should this be here?

andrew-fleming: It should not!

